PR #6, opened 2026-04-07
TRUNK-6608: Add Velero backup and restore as optional feature in OpenMRS Helm chart

praisekavuma4:
Summary
This PR moves the Velero backup and restore solution into the OpenMRS Helm chart so that ALL 
OpenMRS Kubernetes deployments worldwide can benefit from it, not just 
o3-k8s.openmrs.org.

Changes Made

Modified Files
- helm/openmrs/Chart.yaml — bumped chart version from 1.2.2 to 1.2.3 
  and added Velero as an optional dependency that only gets installed 
  when backup is enabled
- helm/openmrs/values.yaml — added a new `backup` section with all 
  configurable options for backup and restore

New Files
- helm/openmrs/templates/velero-backup-schedule.yaml — creates the 
  AWS credentials secret, the backup storage location and the daily 
  backup schedule. Also includes pre-backup hooks that flush MariaDB 
  to a dump file and snapshot Elasticsearch before each backup
- helm/openmrs/templates/velero-restore-cronjob.yaml— runs a health 
  check on the configured schedule and automatically triggers a restore 
  from the latest completed backup if the cluster is found to be 
  genuinely broken

Related
- JIRA: https://openmrs.atlassian.net/browse/TRUNK-6608


praisekavuma4:
@rkorytkowski . check on this

praisekavuma4:
@dkayiwa kindly check on this

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,99 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  cloud: |
+    [default]
+    aws_access_key_id={{ .Values.backup.credentials.accessKey }}
+    aws_secret_access_key={{ .Values.backup.credentials.secretKey }}
+---
```
suubi-joshua, line 17:
```suggestion

```

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,99 @@
...
+# Velero BackupStorageLocation — tells Velero where to store backups (TRUNK-6608)
```
suubi-joshua:
```suggestion

```

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,157 @@
...
+---
+# ServiceAccount — the identity that the restore checker runs 
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}-restore-checker
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+---
```
suubi-joshua:
```suggestion
```

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,157 @@
...
+# CronJob — health check that auto-restores when OpenMRS is genuinely broken (TRUNK-6608)
```
suubi-joshua:
```suggestion
```

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,157 @@
...
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: {{ .Release.Name }}-restore-checker
+          restartPolicy: OnFailure
+          containers:
+            - name: restore-check
+              image: bitnami/kubectl:1.30
```
suubi-joshua:
```suggestion
              image: bitnami/kubectl:1.30-debian-12
```

suubi-joshua: This CronJob container is using `bitnami/kubectl:1.30` which is Alpine/busybox-based. The `date -d` flag for parsing an ISO8601 timestamp string is GNU coreutils only. On busybox `date -d` is not supported for string parsing, it silently returns empty value. Therefore `TRANSITION_EPOCH` is always going to be empty, `AGE` evaluates to the full value of `$NOW`, the threshold check never passes, `SUSTAINED_BROKEN` stays 0, and the auto-restore never triggers.

Switching the container image to a Debian-based variant like I have suggested, this will this will enable GNU coreutils ships with the Debian image, including `date -d` support natively. 

@rkorytkowski i dont know if you have a better approach on this?

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,99 @@
...
+  provider: aws
+  objectStorage:
+    bucket: {{ .Values.backup.bucket }}
+  config:
+    region: {{ .Values.backup.region }}
+    s3ForcePathStyle: "true"
+  credential:
+    name: {{ .Release.Name }}-velero-credentials
+    key: cloud
+---
```
suubi-joshua:
```suggestion
```

helm/openmrs/Chart.yaml
```diff
@@ -33,3 +33,7 @@ dependencies:
     repository: "file://../openmrs-frontend"
     #repository: "oci://ghcr.io/openmrs"
     condition: openmrs-frontend.enabled
+  - name: velero
```
suubi-joshua: We might run into a `race condition` during clean installs here. Because Helm applies all templates in a single transaction, Kubernetes will try to create the `BackupStorageLocation` and `Schedule` resources before the Velero CRDs are fully registered. This might cause the installation to fail immediately.

I would suggest that we remove Velero as a bundled sub-chart and treat it as a cluster-level prerequisite. To prevent 'half-baked' installs where some resources fail while others succeed.

Also Velero is a cluster-wide service; bundling it here means every application update risks an unintended Velero upgrade.

After removing the sub-chart dependency, add a pre-install hook to verify the CRDs exist. This will ensure the install 'fails fast' with a clear instruction if Velero isn't ready.

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,99 @@
...
+                command:
+                  - /bin/sh
+                  - -c
+                  - >
+                    mysqldump
+                    -u root
+                    -p${MARIADB_ROOT_PASSWORD}
+                    --all-databases
+                    --single-transaction
+                    > /bitnami/mariadb/data/full-backup.sql
```
suubi-joshua: The dump writes to `/bitnami/mariadb/data/` this is the same directory MariaDB uses for live InnoDB data files. This is bad because
1. `full-backup.sql` accumulates with every backup. There is no cleanup step, so after backups the data volume will hold stale dump files sitting alongside live data. Over time this wastes significant disk space.
2. Some MariaDB versions log warnings about unexpected files in the data directory on startup, which pollutes logs and could cause operational confusion.

@praisekavuma4 could you replace the Velero pre-hook approach entirely with a dedicated CronJob that runs mysqldump and pipes directly to S3. This is cleaner for three reasons: (1) it decouples DB backup from Velero's hook timing, (2) the dump goes straight to object storage without needing to be on a PVC, and (3) it can run on its own schedule independent of the Velero volume backup.
Remove the mariadb-backup-hook block from velero-backup-schedule.yaml and add a new template.

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,99 @@
...
+        - name: elasticsearch-backup-hook
+          includedNamespaces:
+            - {{ .Release.Namespace }}
+          labelSelector:
+            matchLabels:
+              app.kubernetes.io/name: elasticsearch
+          pre:
+            - exec:
+                container: elasticsearch
+                command:
```
suubi-joshua, line 68: This command creates a snapshot inside an Elasticsearch repository named velero. A snapshot repository must be registered first via `PUT /_snapshot/velero` with a backing store (S3 bucket, path, credentials). The `curl` will produce a failure.

Replace the pre-hook with a dedicated Helm post-install `Job` that registers the snapshot repository once at install time, not on every backup. Running registration inside the pre-hook means it fires on every backup attempt which adds latency and noise in the logs.

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,157 @@
...
+                    --no-headers 2>/dev/null | wc -l)
+
+                  echo "Running pods: $RUNNING"
+
+                  # Find pods that have been in a broken state
+                  # longer than the threshold (default 15 minutes)
+                  # This avoids false triggers during normal pod restarts
+                  SUSTAINED_BROKEN=0
+
+                  for pod in $(kubectl get pods -n "$NAMESPACE" \
```
suubi-joshua: `$NAMESPACE` is the OpenMRS application namespace (e.g. openmrs). Velero stores its Backup and Restore CRDs in its own namespace typically `velero`. These queries will always return empty, causing the script to log ` "No completed backups found. Cannot restore." ` and exit without restoring anything, even when valid backups exist.

You should add a configurable Velero namespace and use it for all CRD queries.

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,157 @@
...
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list"]
+  - apiGroups: ["velero.io"]
+    resources: ["backups", "restores"]
+    verbs: ["get", "list", "create"]
+---
```
suubi-joshua:
```suggestion
```

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,157 @@
...
+# ClusterRoleBinding — connects the permission above to our service account (TRUNK-6608)
```
suubi-joshua:
```suggestion
# ClusterRoleBinding
```

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,157 @@
+{{- if .Values.backup.enabled }}
+# ClusterRole — gives the restore checker permission to read pods and manage restores (TRUNK-6608)
```
suubi-joshua:
```suggestion
```

helm/openmrs/values.yaml
```diff
@@ -40,3 +40,35 @@ openmrs-backend:
...
+    secretKey: ""
+
+  # Restore health check settings.
+  restore:
+    # How often to run the health check.
+    schedule: "*/30 * * * *"
+
+    # How many minutes a pod must be broken before a restore is triggered.
+    # This prevents false triggers during normal pod restarts.
+    brokenThresholdMinutes: 15
```
suubi-joshua: Would it be safer to increase this to 30?

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,99 @@
...
+apiVersion: velero.io/v1
+kind: Schedule
+metadata:
+  name: {{ .Release.Name }}-daily-backup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+spec:
+  schedule: {{ .Values.backup.schedule | quote }}
+  useOwnerReferencesInBackup: false
```
suubi-joshua:
```suggestion
```

suubi-joshua: `useOwnerReferencesInBackup: false` is the Velero default

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,99 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+spec:
+  provider: aws
+  objectStorage:
+    bucket: {{ .Values.backup.bucket }}
+  config:
+    region: {{ .Values.backup.region }}
+    s3ForcePathStyle: "true"
```
suubi-joshua: Path-style S3 access is only needed for `MinIO` and other S3-compatible endpoints, not standard AWS S3. Hardcoding `"true"` for all deployments misconfigures operators using real AWS S3 buckets.
 We should make it configurable with a safe default in `values.yaml`

suubi-joshua (changes requested):
Thanks for working on this @praisekavuma4 I have left some specific reviews kindly look into them. Also after making changes test and see that tests pass.

helm/openmrs/templates/velero-restore-cronjob.yaml
rkorytkowski: It sounds like a good approach.

rkorytkowski: Actually, I don't think auto-restore based on health checks is something we should be supporting. A failure may have nothing to do with data so it should be always admin's action to restore data.

In our demo environment, we will want to be able to just run a restore from a single backup on a schedule, e.g. daily. If the demo is broken, then we may want to trigger an additional restore manually.

suubi-joshua: Okay 100% agree, this changes approach completely

helm/openmrs/templates/velero-backup-schedule.yaml
praisekavuma4: Thank you @suubi-joshua for the insights. let me reconsider👍👍

dkayiwa:
@coderabbitai full review

helm/openmrs/templates/velero-restore-cronjob.yaml
praisekavuma4, line 41: Fixed. The stray --- has been removed from inside the credentials string content.

helm/openmrs/templates/velero-restore-cronjob.yaml
praisekavuma4: Fixed. Updated to bitnami/kubectl:1.30-debian-12 to ensure GNU coreutils
is available for the date -d flag used in timestamp parsing.
Thank you for catching this — on Alpine this would have silently broken
the restore logic entirely.

praisekavuma4: Agreed. The health-check auto-restore approach has been completely removed.
The restore CronJob now runs on a daily schedule and restores from the
latest completed backup. Manual restore remains the admin's responsibility
when something is broken for reasons unrelated to data.

helm/openmrs/templates/velero-backup-schedule.yaml
praisekavuma4: Fixed. The Velero pre-hook approach for MariaDB has been removed entirely.
A dedicated CronJob (mariadb-backup-cronjob.yaml) now runs mysqldump
and pipes the output directly to S3 using a two-container approach —
one container does the dump to a shared volume, the other uploads to S3.
This decouples the DB backup from Velero's hook timing and avoids
writing anything to the live data directory.

helm/openmrs/templates/velero-backup-schedule.yaml
praisekavuma4: Fixed. Removed since it is the Velero default and does not need
to be explicitly stated.

helm/openmrs/templates/velero-backup-schedule.yaml
praisekavuma4: Fixed. s3ForcePathStyle is now configurable in values.yaml with
a default of false, which is correct for standard AWS S3.
Users running MinIO or other S3-compatible endpoints can set
it to true in their own values file.

praisekavuma4:
@rkorytkowski , @suubi-joshua , @dkayiwa kindly check on the new fixes. thank you

praisekavuma4:
https://github.com/coderabbitai full review

suubi-joshua (changes requested):
@praisekavuma4. I have reviewed the current state of the PR including the latest fix commit (c32ce4e). Below are the remaining issues that need attention.

**1. Missing `_helpers.tpl`,  undefined template `openmrs.labels`**

Every new template (`mariadb-backup-cronjob.yaml:9,25`, `velero-backup-schedule.yaml:10,25,44`, `velero-restore-cronjob.yaml:7,22,39,49`) references `{{- include "openmrs.labels" . }}`, but the parent chart at `helm/openmrs/templates/` has **no `_helpers.tpl`** file. The sub-charts define `openmrs-backend.labels` and `openmrs-frontend.labels`, but `openmrs.labels` does not exist anywhere. `helm install` will fail with:

```
Error: template "openmrs.labels" not defined
```

**Fix:** Create `helm/openmrs/templates/_helpers.tpl` following the same pattern as the sub-charts, e.g.:
```yaml
{{- define "openmrs.labels" -}}
helm.sh/chart: {{ include "openmrs.chart" . }}
{{ include "openmrs.selectorLabels" . }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}

{{- define "openmrs.selectorLabels" -}}
app.kubernetes.io/name: {{ include "openmrs.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end }}

{{- define "openmrs.name" -}}
{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
{{- end }}

{{- define "openmrs.chart" -}}
{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
{{- end }}
```

**2. Dangling `fi` in restore-check shell script**

`velero-restore-cronjob.yaml` line 113 has a trailing `fi` with no matching `if`. The shell script already has two `if` blocks — both properly closed — leaving this third `fi` orphaned. This will cause a shell syntax error when the CronJob runs.

**Fix:** Delete line 113 (`                  fi`).

**3. `--no-sign-request` flag contradicts credentials in MariaDB backup**

`mariadb-backup-cronjob.yaml:76-78`:
```yaml
{{- if .Values.backup.mariadbBackup.s3ForcePathStyle }}
--no-sign-request \
{{- end }}
```

`--no-sign-request` makes **unsigned anonymous requests**, which directly contradicts the AWS credentials provided via `envFrom` on the same container. This flag is for public/anonymous bucket access, not for S3-compatible storage. When using MinIO (the primary use case for `s3ForcePathStyle: true`), you need `--endpoint-url <url>`, not `--no-sign-request`.

**Fix:** Replace with an `--endpoint-url` parameter:
```yaml
{{- if .Values.backup.mariadbBackup.s3Endpoint }}
--endpoint-url {{ .Values.backup.mariadbBackup.s3Endpoint }} \
{{- end }}
```
And add `s3Endpoint: ""` to `values.yaml` under `backup.mariadbBackup`.

**4. Elasticsearch snapshot repository not pre-registered**

`velero-backup-schedule.yaml:68-71` does:
```bash
curl -s -X PUT "localhost:9200/_snapshot/velero/snapshot_$(date +%Y%m%d%H%M)?wait_for_completion=true"
```

The `velero` snapshot repository must be registered first via `PUT _snapshot/velero` with a backing store configuration. Without prior registration, this curl call will return a 400 error, and since `onError: Fail` is set, the entire Velero backup will fail.

**Fix:** Either add a one-time Job to register the repo at install time, or change the pre-hook to include registration logic:
```yaml
command:
  - /bin/sh
  - -c
  - >
    curl -s -X PUT "localhost:9200/_snapshot/velero" -H "Content-Type: application/json" -d {"type": "fs", "settings": {"location": "/usr/share/elasticsearch/data/snapshots"}} &&
    curl -s -X PUT "localhost:9200/_snapshot/velero/snapshot_$(date +%Y%m%d%H%M)?wait_for_completion=true"
```

(Adjust the repository type/location as appropriate for your setup note: this is the default Elasticsearch fs snapshot repo.)

**5. MariaDB root password exposed on command line**

`mariadb-backup-cronjob.yaml:53-54`:
```bash
mysqldump -h ... -u root -p${MARIADB_ROOT_PASSWORD} ...
```

The password is passed as a command-line argument, making it visible via `/proc` to other containers and in process listings. Use `MYSQL_PWD` environment variable instead:

```bash
MYSQL_PWD="${MARIADB_ROOT_PASSWORD}" mysqldump -h ... -u root --all-databases --single-transaction > /dump/backup.sql
```

**6. Hardcoded MariaDB service names**

`mariadb-backup-cronjob.yaml:44,52` references `{{ .Release.Name }}-mariadb` and `{{ .Release.Name }}-mariadb-primary`. If someone sets `nameOverride` or `fullnameOverride` on the MariaDB sub-chart, these will be wrong. Make these configurable via values.

**7. No `required` validation for empty bucket/credentials**

`velero-backup-schedule.yaml:15-16,29` and `mariadb-backup-cronjob.yaml:12-13` will silently template empty strings for `bucket`, `accessKey`, and `secretKey` when `backup.enabled: true`. This causes a misconfiguration that only surfaces at runtime inside Velero. Use Helm's `required` function:

```yaml
bucket: {{ required "backup.bucket is required when backup.enabled=true" .Values.backup.bucket }}
```

praisekavuma4:
thank u  @suubi-joshua. for the review

suubi-joshua:
HI @praisekavuma4 still working on this?

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,114 @@
...
+                    namespace: $VELERO_NAMESPACE
+                  spec:
+                    backupName: $LATEST_BACKUP
+                    includedNamespaces:
+                      - $APP_NAMESPACE
+                    existingResourcePolicy: replace
+                  EOF
+
+                  echo "Restore submitted successfully."
+                  fi
```
suubi-joshua: This `fi` has no matching `if` block. The shell script properly closes both if statements on lines 79–82 and 92–95, making this third `fi` a syntax error that will crash the CronJob at runtime.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,83 @@
...
+                    name: {{ .Release.Name }}-mariadb-backup-credentials
+              command:
+                - /bin/sh
+                - -c
+                - |
+                  echo "Uploading dump to S3..."
+                  aws s3 cp /dump/backup.sql \
+                    s3://{{ .Values.backup.mariadbBackup.bucket }}/mariadb-backup-$(date +%Y%m%d%H%M).sql \
+                    --region {{ .Values.backup.mariadbBackup.region }} \
+                    {{- if .Values.backup.mariadbBackup.s3ForcePathStyle }}
```
suubi-joshua: The code uses --no-sign-request when s3ForcePathStyle: true, but this is contradictory:

- `--no-sign-request` makes unsigned anonymous requests
- `envFrom` supplies AWS credentials that should be signed
- For MinIO or other S3-compatible endpoints, you need `--endpoint-url <url>`, not `--no-sign-request`

This will cause authentication failures when using MinIO.

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,74 @@
...
+          labelSelector:
+            matchLabels:
+              app.kubernetes.io/name: elasticsearch
+          pre:
+            - exec:
+                container: elasticsearch
+                command:
+                  - /bin/sh
+                  - -c
+                  - >
```
suubi-joshua, line 71: The pre-hook attempts to take a snapshot at `PUT _snapshot/velero/snapshot_<timestamp>`, but the `velero` repository must first be registered via `PUT _snapshot/velero` with storage configuration. Without this, Elasticsearch returns a 400 error and the backup fails (since `onError: Fail`).

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,83 @@
...
+                      key: mariadb-root-password
+              command:
+                - /bin/sh
+                - -c
+                - |
+                  echo "Starting MariaDB dump..."
+                  mysqldump \
+                    -h {{ .Release.Name }}-mariadb-primary \
+                    -u root \
+                    -p${MARIADB_ROOT_PASSWORD} \
```
suubi-joshua: The password is passed as -`p${MARIADB_ROOT_PASSWORD}`, making it visible to other containers on the same node via /proc and in process listings.

The `MYSQL_PWD` environment variable is the MySQL-recommended way and hides the password from process listings.

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,74 @@
...
+metadata:
+  name: {{ .Release.Name }}-velero-credentials
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  cloud: |
+    [default]
+    aws_access_key_id={{ .Values.backup.credentials.accessKey }}
```
suubi-joshua: When `backup.enabled=true`, empty strings for bucket, `accessKey`, and `secretKey` silently template into the `YAML`, creating a misconfigured Velero that only fails at runtime

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,83 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-mariadb-backup-credentials
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  AWS_ACCESS_KEY_ID: {{ .Values.backup.mariadbBackup.credentials.accessKey | quote }}
```
suubi-joshua: When `backup.enabled=true`, empty strings for bucket, `accessKey`, and `secretKey` silently template into the YAML, creating a misconfigured Velero that only fails at runtime

Add Helm's `required()` function

suubi-joshua:
The `velero-restore-cronjob.yaml` unconditionally restores from the latest backup daily with `existingResourcePolicy: replace`, which overwrites the live namespace, destroying any changes made since the last backup. This is not industry standard; [Velero docs](https://velero.io/docs/v1.18/restore-reference/) recommend restore as an on-demand, admin-triggered action.

- Remove auto-restore entirely and document manual restore procedure
- Add health-check gates (readiness probes must fail N times before restore triggers)
- Make it opt-in with backup.restore.enabled: false by default

praisekavuma4:
> HI @praisekavuma4 still working on this?

yes back to accomplish it

helm/openmrs/templates/velero-restore-cronjob.yaml
praisekavuma4: Fixed. Removed the trailing fi that had no matching if block.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
praisekavuma4: Fixed. Replaced -p${MARIADB_ROOT_PASSWORD} with the MYSQL_PWD
environment variable which hides the password from process listings
and /proc as recommended.

praisekavuma4:
> The `velero-restore-cronjob.yaml` unconditionally restores from the latest backup daily with `existingResourcePolicy: replace`, which overwrites the live namespace, destroying any changes made since the last backup. This is not industry standard; [Velero docs](https://velero.io/docs/v1.18/restore-reference/) recommend restore as an on-demand, admin-triggered action.
> 
> * Remove auto-restore entirely and document manual restore procedure
> * Add health-check gates (readiness probes must fail N times before restore triggers)
> * Make it opt-in with backup.restore.enabled: false by default

The restore CronJob is now opt-in and disabled by default
via backup.restore.enabled: false in values.yaml. Administrators
who want scheduled restores for demo environments can explicitly
enable it. Manual restore remains the recommended approach as per
Velero documentation.

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,113 @@
...
+                  apiVersion: velero.io/v1
+                  kind: Restore
+                  metadata:
+                    name: scheduled-restore-$(date +%Y%m%d%H%M)
+                    namespace: $VELERO_NAMESPACE
+                  spec:
+                    backupName: $LATEST_BACKUP
+                    includedNamespaces:
+                      - $APP_NAMESPACE
+                    existingResourcePolicy: replace
```
suubi-joshua, line 113: `replace` is not a valid value in Velero v1.18 (Restore API Type (https://velero.io/docs/v1.18/api-types/restore/)).

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,78 @@
...
+type: Opaque
+stringData:
+  cloud: |
+    [default]
+    aws_access_key_id={{ required "backup.credentials.accessKey is required when backup.enabled=true" .Values.backup.credentials.accessKey }}
+    aws_secret_access_key={{ required "backup.credentials.secretKey is required when backup.enabled=true" .Values.backup.credentials.secretKey }}
+---
+# Velero BackupStorageLocation — tells Velero where to store backups 
+apiVersion: velero.io/v1
+kind: BackupStorageLocation
```
suubi-joshua, line 20: Move BSL/Schedule to Velero namespace, Change `namespace: {{ .Release.Namespace }}` to `namespace: {{ .Values.backup.veleroNamespace }}` on the Secret (line 8), BackupStorageLocation (line 23), and Schedule (line 42) in velero-backup-schedule.yaml. Correct me if am wrong but I think Velero watches its own namespace by default, not the app namespace

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,113 @@
+{{- if and .Values.backup.enabled .Values.backup.restore.enabled }}
```
suubi-joshua, line 1: Shouldn't it be `{{- if and .Values.backup.enabled .Values.backup.restore.enabled }}`

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,78 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+spec:
+  provider: aws
+  objectStorage:
+    bucket: {{ required "backup.bucket is required when backup.enabled=true" .Values.backup.bucket }}
+  config:
+    region: {{ .Values.backup.region }}
+    s3ForcePathStyle: {{ .Values.backup.s3ForcePathStyle | quote }}
```
suubi-joshua, line 32: We have no way to set the S3 endpoint for MinIO:
Should we add:
   ` {{- if .Values.backup.s3Url }}
    s3Url: {{ .Values.backup.s3Url }}
    {{- end }}`

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,78 @@
+{{- if .Values.backup.enabled }}
+# Velero backup credentials secret 
+# Stores AWS credentials securely as a Kubernetes secret
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-velero-credentials
+  namespace: {{ .Release.Namespace }}
```
suubi-joshua: Velero only watches its own namespace
I think it Should be:
  `namespace: {{ .Values.backup.veleroNamespace }}`

suubi-joshua (commented):
Also as could you rebase with main.

helm/openmrs/values.yaml
```diff
@@ -40,3 +40,65 @@ openmrs-backend:
...
+  # MariaDB backup settings — runs a dedicated CronJob that dumps the DB directly to S3.
+  mariadbBackup:
+    enabled: false
+    schedule: "30 1 * * *"
+    bucket: ""
+    region: "us-east-1"
+    s3ForcePathStyle: false
+
+    # S3-compatible endpoint URL. Only needed for MinIO or other non-AWS endpoints.
+    # Example: "http://minio.minio-namespace.svc.cluster.local:9000"
```
suubi-joshua, line 100: Since this chart's environment uses SeaweedFS S3, these should reference the SeaweedFS S3 service instead. Also, the chart should include a setup job or init hook that creates the backup bucket in SeaweedFS S3 (openmrs-backups), similar to how the MinIO test setup had a minio-setup job.

helm/openmrs/templates/velero-restore-cronjob.yaml
praisekavuma4: @suubi-joshua  this is exactly the same. it was fixed in the latest commit.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,82 @@
...
+                - -c
+                - |
+                  echo "Uploading dump to S3..."
+                  aws s3 cp /dump/backup.sql \
+                    s3://{{ required "backup.mariadbBackup.bucket is required when backup.mariadbBackup.enabled=true" .Values.backup.mariadbBackup.bucket }}/mariadb-backup-$(date +%Y%m%d%H%M).sql \
+                    --region {{ .Values.backup.mariadbBackup.region }} \
+                    {{- if .Values.backup.mariadbBackup.s3Endpoint }}
+                    --endpoint-url {{ .Values.backup.mariadbBackup.s3Endpoint }} \
+                    {{- end }}
+                  echo "Upload complete."
```
dkayiwa, line 84: This upload never actually runs, so with `mariadbBackup.enabled=true` the dump is taken but nothing reaches S3. The last argument line ends with a trailing `\`, so the shell folds `echo "Upload complete."` into the `aws s3 cp` command and passes `echo` and `Upload complete.` to it as extra positional arguments, and `aws s3 cp` then exits with a usage error. I rendered the template and ran the result to confirm it (aws gets invoked with `echo` and `Upload complete.` as trailing args). It breaks with or without `s3Endpoint`, because the line before `echo` always ends in `\`: the `--endpoint-url` line when the endpoint is set, the `--region` line when it isn't. The pod exits non-zero, so `restartPolicy: OnFailure` re-dumps and re-fails on a loop, and it fails quietly with nothing in S3.

Pulling the endpoint into a shell variable keeps the line continuations correct either way:

```suggestion
                  ENDPOINT_ARG=""
                  {{- if .Values.backup.mariadbBackup.s3Endpoint }}
                  ENDPOINT_ARG="--endpoint-url {{ .Values.backup.mariadbBackup.s3Endpoint }}"
                  {{- end }}
                  aws s3 cp /dump/backup.sql \
                    s3://{{ required "backup.mariadbBackup.bucket is required when backup.mariadbBackup.enabled=true" .Values.backup.mariadbBackup.bucket }}/mariadb-backup-$(date +%Y%m%d%H%M).sql \
                    --region {{ .Values.backup.mariadbBackup.region }} $ENDPOINT_ARG
                  echo "Upload complete."
```

(Separately, `mariadbBackup.s3ForcePathStyle` in values.yaml isn't referenced by this template, so it can't actually be turned on. The aws CLI generally needs path-style addressing for MinIO, so that upload would still fail for MinIO even after the fix above.)

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,82 @@
...
+              emptyDir: {}
+          initContainers:
+            - name: dump
+              image: bitnami/mariadb:10.11
+              env:
+                - name: MARIADB_ROOT_PASSWORD
+                  valueFrom:
+                    secretKeyRef:
+                      name: {{ .Release.Name }}-mariadb
+                      key: mariadb-root-password
```
dkayiwa: This secret reference doesn't match how MariaDB is deployed here, so the dump init container can't start. It points at secret `{{ .Release.Name }}-mariadb` / key `mariadb-root-password`, which is the Bitnami chart layout, but MariaDB in this chart comes from mariadb-operator and the root password lives in `{{ include "openmrs-backend.fullname" . }}-mariadb-secret` under key `root-password` (see helm/openmrs-backend/templates/mariadb-secret.yaml). With `mariadbBackup.enabled=true` the `secretKeyRef` resolves to a secret and key that don't exist, so the pod stays in CreateContainerConfigError and no dump is ever produced. I confirmed the names by rendering both charts.

With the backend `nameOverride` set to `backend`, that secret renders as `<release>-backend-mariadb-secret`:

```suggestion
                      name: {{ .Release.Name }}-backend-mariadb-secret
                      key: root-password
```

That hardcodes the `backend` nameOverride, the same way the host default already assumes the operator's service name. If you'd rather not depend on it, exposing the secret name and key as values under `mariadbBackup` would be more robust.

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,113 @@
...
+# ServiceAccount — the identity that the restore checker runs 
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}-restore-checker
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs.labels" . | nindent 4 }}
+---
+# CronJob — health check that auto-restores when OpenMRS is genuinely broken
```
dkayiwa: Now that the health-check logic is gone (per the earlier thread), this comment and the `restore-check` / `restore-checker` names describe behavior that no longer exists. The job doesn't check anything or detect breakage, it restores from the latest backup on every scheduled run, which is what was requested. The risk is that someone reading `restore-check` or "auto-restores when OpenMRS is genuinely broken" assumes it only fires on failure and enables it as a safe guard, when it actually replaces the namespace on a daily schedule. I'd rename it to something like `scheduled-restore` and reword the comment to say it's an unconditional scheduled restore. The PR description still describes the old health-check version too.

helm/openmrs/templates/velero-crd-check.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.backup.enabled }}
+{{- /*
+  Pre-install check — verifies Velero CRDs are registered before
+  creating backup resources
+  This check is skipped automatically during helm template (offline)
+  since .Capabilities returns empty without a cluster connection.
```
dkayiwa: This comment isn't accurate, and it matters for testing. `helm template` doesn't get empty Capabilities; Helm fills in a default API list, so `.Capabilities.APIVersions` is non-empty and the `else if` branch fires. `helm template --set backup.enabled=true` fails with the "Velero is not installed" error unless you also pass `--api-versions velero.io/v1/Schedule` (I hit this rendering the backup templates). The effect is that none of the backup manifests can be rendered or linted offline, which is likely why the template issues in this PR went unnoticed, since CI only renders with `backup.enabled=false`. Worth either correcting the comment or letting the check pass when the API list is the offline default, so the manifests can be dry-run rendered.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
praisekavuma4: @dkayiwa , I fixed this in the latest commit 
using the shell variable approach you suggested.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,88 @@
...
+                - name: dump
+                  mountPath: /dump
+          containers:
+            - name: upload
+              image: amazon/aws-cli:2.15.0
+              envFrom:
+                - secretRef:
+                    name: {{ .Release.Name }}-mariadb-backup-credentials
+              command:
+                command:
```
dkayiwa, line 68: This stray `command:` key blocks merge. It turns the container's `command` into a map instead of a list, so with `mariadbBackup.enabled=true` the API server rejects the manifest and the whole `helm install`/`helm upgrade` fails, not just the backup resources. I rendered the chart with helm 3.17.4 (the version CI pins) and validated the output against the v1.30 CronJob schema: it fails at `spec.jobTemplate.spec.template.spec.containers[0].command` with `{'command': ['/bin/sh', '-c', ...]} is not of type 'array', 'null'`. `Container.Command` is `[]string` in `k8s.io/api/core/v1`, so there is no form of this the API server accepts.

```suggestion
              command:
```

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,88 @@
...
+                - -c
+                - |
+                  echo "Starting MariaDB dump..."
+                  MYSQL_PWD="${MARIADB_ROOT_PASSWORD}" mysqldump \
+                    -h {{ default (printf "%s-mariadb-primary" .Release.Name) .Values.backup.mariadbBackup.mariadbHost }} \
+                    -u root \
+                    --all-databases \
+                    --single-transaction \
+                    > /dump/backup.sql
+                  echo "Dump complete. Size: $(du -h /dump/backup.sql | cut -f1)"
```
dkayiwa, line 57: Both containers here report success when they fail, and I'd want that fixed before this merges. Neither script sets `-e`, and in each one the last command is an `echo`, so the script's exit status is the echo's rather than that of the command doing the work.

I ran the rendered scripts with a stubbed `mysqldump` that exits 2. The redirect still creates `/dump/backup.sql`, the script prints `Dump complete. Size: 0B` and exits 0, the init container is marked Succeeded, and the upload container then copies the empty file up. Repeating it with a failing `aws s3 cp` prints `Upload complete.` and also exits 0. If merged as-is, a wrong DB host or a rotated S3 key gives you a CronJob whose history is entirely Completed plus a bucket of 0-byte objects, and the first thing that tells you is a restore with nothing in it.

`set -e` is enough, since the trailing `echo` then never runs (checked with the same stub: exit 2, no "Dump complete" line):

```suggestion
                  set -e
                  echo "Starting MariaDB dump..."
                  MYSQL_PWD="${MARIADB_ROOT_PASSWORD}" mysqldump \
                    -h {{ default (printf "%s-mariadb-primary" .Release.Name) .Values.backup.mariadbBackup.mariadbHost }} \
                    -u root \
                    --all-databases \
                    --single-transaction \
                    > /dump/backup.sql
                  echo "Dump complete. Size: $(du -h /dump/backup.sql | cut -f1)"
```

The upload script needs the same treatment at line 84, and so does `velero-bucket-setup.yaml` line 39, where `2>&1 | grep -v "BucketAlreadyOwnedByYou" || true` swallows connection and auth errors along with the one it means to ignore, so that Job reports success without having created the bucket.

helm/openmrs/templates/velero-backup-schedule.yaml
```diff
@@ -0,0 +1,81 @@
...
+    hooks:
+      resources:
+        {{- if (index .Values "openmrs-backend" "elasticsearch" "enabled") }}
+        # Pre-backup hook — snapshot Elasticsearch before backup starts
+        - name: elasticsearch-backup-hook
+          includedNamespaces:
+            - {{ .Release.Namespace }}
+          labelSelector:
+            matchLabels:
+              app.kubernetes.io/name: elasticsearch
```
dkayiwa, line 64: This selector does not match the Elasticsearch pods the chart creates, so the pre-backup hook never runs. Elasticsearch here comes from the `eck-elasticsearch` subchart, and ECK builds its pod labels in `NewPodLabels` (`pkg/controller/elasticsearch/label/label.go`): `common.k8s.elastic.co/type: elasticsearch`, `elasticsearch.k8s.elastic.co/cluster-name`, the version and node-role labels, and the statefulset name. No `app.kubernetes.io/*` label is among them, and the podTemplate this chart renders adds no labels of its own. The one `app.kubernetes.io/name` in the picture sits on the `Elasticsearch` CR itself, with value `elasticsearch-eck`.

If merged as-is, no pod matches, so Velero copies the Elasticsearch data directory live with no flush or snapshot, and since there is no hook execution to fail, `onError: Fail` never comes into play and the Backup still finishes Completed.

```suggestion
          labelSelector:
            matchLabels:
              common.k8s.elastic.co/type: elasticsearch
```

`kind-openmrs.yaml` sets `elasticsearch.enabled: true`, so this is the path the full-featured config takes.

helm/openmrs/templates/velero-bucket-setup.yaml
```diff
@@ -0,0 +1,42 @@
...
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: create-bucket
+          image: amazon/aws-cli:2.15.0
+          env:
+            - name: AWS_ACCESS_KEY_ID
+              value: {{ .Values.backup.credentials.accessKey | quote }}
+            - name: AWS_SECRET_ACCESS_KEY
+              value: {{ .Values.backup.credentials.secretKey | quote }}
```
dkayiwa, line 29: These two should come from a Secret before this merges. An env `value:` is stored inline in the pod spec, so the S3 secret key becomes readable by anyone who can run `kubectl get job -o yaml` or `kubectl describe pod` in the release namespace, a considerably wider group than the people who can read Secrets. I rendered with `backup.credentials.secretKey=SUPERSECRET` and it comes back as `{'name': 'AWS_SECRET_ACCESS_KEY', 'value': 'SUPERSECRET'}` in the Job spec.

`mariadb-backup-cronjob.yaml` already does this the right way with `envFrom`/`secretRef`. The `-velero-credentials` Secret cannot be reused here because it lands in `backup.veleroNamespace` while this pod runs in the release namespace, so the smallest fix is a small Secret alongside this Job holding `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`, with `envFrom` on the container.

helm/openmrs/Chart.yaml
```diff
@@ -33,3 +33,4 @@ dependencies:
     repository: "file://../openmrs-frontend"
     #repository: "oci://ghcr.io/openmrs"
     condition: openmrs-frontend.enabled
+
```
dkayiwa, line 36: This blank line is all that is left of the `Chart.yaml` change, and it is only trailing whitespace, so it can go.

While you are in here, the PR description has drifted from the change. It says `Chart.yaml` was bumped 1.2.2 to 1.2.3 and gained Velero as an optional dependency, neither of which is in the diff (and treating Velero as a cluster prerequisite instead looks like the deliberate outcome of the earlier discussion). It also still describes the restore CronJob as health-checking and restoring only when the cluster is found to be broken.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
dkayiwa: This is back at line 79 of the current head: `mariadbBackup.s3ForcePathStyle: true` now sets `PATH_STYLE_ARG="--no-sign-request"`, wired up in the latest commit.

Your reading still holds, and there is no environment-variable route to path-style either. In botocore's `DEFAULT_S3_CONFIG_VARS` (`botocore/configprovider.py`), `addressing_style` is declared as `(('s3', 'addressing_style'), None, None, None)`; the second slot is the env-var name and it is `None`, unlike neighbours such as `s3_disable_express_session_auth` which do carry one. So it is config-file only, and `aws configure set default.s3.addressing_style path` before the copy is the way to get it. `--no-sign-request` should just come out.

If merged as-is, an operator who sets `s3ForcePathStyle: true`, which is exactly what SeaweedFS and MinIO need and SeaweedFS is what this repo runs, gets an unsigned anonymous PutObject that 403s against an authenticated endpoint, while still not getting the path-style addressing the value name promises.

helm/openmrs/templates/velero-backup-schedule.yaml
dkayiwa: The Secret moved to `backup.veleroNamespace`, but the `BackupStorageLocation` (line 23) and the `Schedule` (line 45) are still on `{{ .Release.Namespace }}`, so this is still open.

Your reading of Velero is right, and it is explicit in the server code. `pkg/cmd/server/server.go` builds the controller-runtime manager with `Cache: cache.Options{DefaultNamespaces: map[string]cache.Config{f.Namespace(): {}}}`, and the field comment on `crClient` in the same file spells out that "the controller-manager's client is limited to list namespaced-scoped resources in the namespace where Velero is installed, or the cluster-scoped resources". `NewScheduleReconciler` is likewise constructed with `s.namespace`.

If merged as-is, with `backup.enabled=true` the install succeeds and both objects sit in the app namespace where nothing reconciles them, so no Backup is ever created and `kubectl get schedule -n velero` comes back empty. Nothing errors, which is the part that will cost somebody a real recovery. Rendering with `--set backup.enabled=true` confirms where they land: Secret in `velero`, BSL and Schedule in the release namespace.

Moving both to `{{ .Values.backup.veleroNamespace }}` also fixes the `credential` reference on line 37, which Velero resolves within the BSL's own namespace and which therefore currently points at a Secret that is not there.

helm/openmrs/templates/velero-restore-cronjob.yaml
dkayiwa: Still `replace` at line 113. I confirmed it against Velero's source rather than the docs page: `pkg/apis/velero/v1/restore_types.go` defines only `PolicyTypeNone = "none"` and `PolicyTypeUpdate = "update"`, and there is no kubebuilder enum marker on the field, so the API server accepts the object and the check happens later. `pkg/controller/restore_controller.go` around line 364 appends `Invalid ExistingResourcePolicy: replace` to `status.validationErrors`.

If merged as-is, `kubectl apply` succeeds, the CronJob logs "Restore submitted successfully." and exits 0, and the Restore sits in FailedValidation having restored nothing. `update` is the value that matches the intent here.

Worth landing the namespace fix on `velero-backup-schedule.yaml` before this one, though: until the Schedule sits in Velero's namespace there are no Backups at all, so this job always takes the "No completed backups found" branch and neither the bug nor its fix is observable.

helm/openmrs/templates/velero-crd-check.yaml
dkayiwa: The NOTE at the top reads correctly now, but the comment at lines 17-21 and the guard it describes do not hold. `.Capabilities.KubeVersion.Version` is never empty during `helm template`; Helm fills in a default, which on 3.17.4 is `v1.32.0`, so `empty` returns false, `not (empty ...)` is always true, and the `else if` behaves as a plain `else`. Rendering with the CI-pinned 3.17.4 still fails: `helm template helm/openmrs --set backup.enabled=true` gives `Velero is not installed on this cluster` unless `--api-versions velero.io/v1/Schedule` is passed too. The file now says both that you need the flag and that you do not.

The coverage gap behind it is the part I would most want closed. Every `Validate Rendering` step in `helm-build.yaml` leaves `backup.enabled` at false, so none of the backup templates are rendered by CI at all, and while that holds the next defect in them ships green.

A render step on its own would not have caught this round's problem, which is more than I allowed for last time. `helm template` renders the stray `command:` key in `mariadb-backup-cronjob.yaml` quite happily; it is the API server, not Helm, that rejects a map where a list belongs. Catching that class needs schema validation of the rendered output. I validated the rendered manifests directly against the standard Kubernetes JSON schemas and the CronJob is the one failure, so `kubeconform`, which reads that same schema set, would gate it:

```
      - name: Validate Rendering (backup enabled)
        run: |
          helm template helm/openmrs --generate-name \
            --api-versions velero.io/v1/Schedule \
            --set backup.enabled=true \
            --set backup.restore.enabled=true \
            --set backup.mariadbBackup.enabled=true \
            --set backup.bucket=test --set backup.mariadbBackup.bucket=test \
            --set backup.credentials.accessKey=x --set backup.credentials.secretKey=y \
            --set backup.mariadbBackup.credentials.accessKey=x \
            --set backup.mariadbBackup.credentials.secretKey=y \
            | kubeconform -strict -summary -ignore-missing-schemas -
```

`-ignore-missing-schemas` is needed because the Velero CRDs have no published JSON schema; the CronJobs, Job, Secrets and RBAC objects all get checked. Dropping the dead `else if` down to a plain `else` at the same time keeps the file honest, since the NOTE already covers the flag.

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,117 @@
+{{- if and .Values.backup.enabled .Values.backup.restore.enabled }}
```
dkayiwa, line 1: This blocks merge, and it takes down every install of the chart, not just the ones with backup enabled. `.Values.backup` is gone: the merge in 3c022a4 took main's `values.yaml` wholesale, and the `backup:` block that 199f1fd added went with it. So `.Values.backup.enabled` on this line dereferences nil.

With helm 3.17.4 (the version CI pins), default values, nothing set:

```
Error: template: openmrs/templates/velero-restore-cronjob.yaml:1:18: executing
"openmrs/templates/velero-restore-cronjob.yaml" at <.Values.backup.enabled>:
nil pointer evaluating interface {}.enabled
```

`origin/main` renders 16 manifests from the same command. This branch renders none, so `helm install` and `helm upgrade` both fail before anything reaches the cluster.

Re-adding the block from 199f1fd is the whole fix. I checked it against the templates: the five files reference 21 distinct `.Values.backup.*` keys and that block defines all 21, so there is nothing new to invent. With it restored I get 16 manifests on defaults and 27 with `--set backup.enabled=true --api-versions velero.io/v1/Schedule`.

Worth knowing why this got this far. The Helm Build run on this head (31796986094) has conclusion `action_required` and zero jobs, so it never executed; the last run that actually ran was on 199f1fd, which still had the block. Nothing since the merge has been through CI.

Two things to pick up on the same trip through `values.yaml`: the PR description still says it gained a `backup` section, which is the claim that made this easy to miss, and the README parameter table has no `backup.*` rows, which TRUNK-6608 lists under acceptance criteria.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,89 @@
...
+    spec:
+      template:
+        spec:
+          restartPolicy: OnFailure
+          volumes:
+            - name: dump
+              emptyDir: {}
+          initContainers:
+            - name: dump
+              image: bitnami/mariadb:10.11
```
dkayiwa, line 39: Neither of the images these two CronJobs pull exists any more, so with `mariadbBackup.enabled=true` and `restore.enabled=true` both pods sit in ImagePullBackOff: no dump reaches S3, no restore is ever submitted, and the only signal is a CronJob that never produces a completed Job.

`docker manifest inspect` returns `no such manifest` for both `bitnami/mariadb:10.11` here and `bitnami/kubectl:1.30-debian-12` in `velero-restore-cronjob.yaml` line 65, and the registry API gives 404 for each. It is the tags, not the repos: `bitnami/kubectl:latest` still resolves 200, and both tags are present under `bitnamilegacy/`. That is Bitnami's move of versioned tags out of the `bitnami/` namespace on Docker Hub.

For the dump, the chart already runs MariaDB from the official image (`mariadb-instance.yaml` line 10), so matching it keeps the version in one place. I pulled `mariadb:10.11` and `/usr/bin/mysqldump` is there (10.11.16-MariaDB):

```suggestion
              image: mariadb:{{ .Values.mariadb.image.tag | default "10.11" }}
```

For the restore job, `registry.k8s.io/kubectl:v1.30.14` is the upstream-published image and resolves fine. The `-debian-12` part has nothing left to do either: since the health-check logic came out, that script only calls `date` and `date +%Y%m%d%H%M`, so the GNU coreutils reason from the earlier thread no longer applies.

Given these tags evidently move under you, I would put both behind values (`backup.mariadbBackup.image`, `backup.restore.image`) so they can be repointed without a chart change.

helm/openmrs/templates/velero-restore-cronjob.yaml
```diff
@@ -0,0 +1,117 @@
...
+                    name: scheduled-restore-$(date +%Y%m%d%H%M)
+                    namespace: $VELERO_NAMESPACE
+                  spec:
+                    backupName: $LATEST_BACKUP
+                    includedNamespaces:
+                      - $APP_NAMESPACE
+                    existingResourcePolicy: replace
+                  EOF
+
+                  echo "Restore submitted successfully."
```
dkayiwa, line 116: Velero resolves "the latest completed backup from this schedule" itself, and its version does not have the defect this pipeline has, so I would hand the job over to it before this merges.

`RestoreSpec.ScheduleName` is documented in `pkg/apis/velero/v1/restore_types.go` as "If specified, and BackupName is empty, Velero will restore from the most recent successful backup created from this schedule", and `pkg/controller/restore_controller.go` (v1.18.2, from line 369) implements it by listing backups with `velero.io/schedule-name == scheduleName` and taking `mostRecentCompletedBackup`. `FromSchedule` in `pkg/builder/backup_builder.go` stamps that label on every scheduled backup, so it is always there to select on.

The difference that matters: `kubectl get backup.velero.io -n velero` returns every backup in Velero's namespace from every schedule in the cluster, so `grep 'Completed' | tail -1` picks the newest backup cluster-wide, not the newest one from this release's Schedule. Velero is a cluster-wide service, so one more OpenMRS release or any unrelated schedule is enough for this job to submit a Restore built from a backup that does not contain `$APP_NAMESPACE`, which restores nothing while logging "Restore submitted successfully." Letting Velero pick also turns the no-backups case into a validation error on the Restore object rather than a shell `exit 0` nobody reads.

Separately, this script never got `set -e`. With a stubbed `kubectl` whose `apply` exits 1 I get the error printed, then "Restore submitted successfully.", then exit 0, so the Job history shows Completed.

```suggestion
                  # Skip if a restore is already running
                  ACTIVE=$(kubectl get restore.velero.io \
                    -n "$VELERO_NAMESPACE" \
                    --no-headers 2>/dev/null | \
                    grep -cE 'InProgress' || true)

                  if [ "$ACTIVE" -gt 0 ]; then
                    echo "A restore is already in progress. Skipping."
                    exit 0
                  fi

                  set -e

                  echo "Submitting restore from schedule {{ .Release.Name }}-daily-backup"

                  cat <<EOF | kubectl apply -f -
                  apiVersion: velero.io/v1
                  kind: Restore
                  metadata:
                    name: scheduled-restore-$(date +%Y%m%d%H%M)
                    namespace: $VELERO_NAMESPACE
                  spec:
                    scheduleName: {{ .Release.Name }}-daily-backup
                    includedNamespaces:
                      - $APP_NAMESPACE
                    existingResourcePolicy: update
                  EOF

                  echo "Restore submitted."
```

I rendered and ran that: shell syntax checks out, and with the failing `kubectl` stub it exits 1. It also makes the `pods` rule and `backups` dead in the ClusterRole above, so `rules` can come down to `get`/`list`/`create` on `restores` alone. `existingResourcePolicy` is `update` in there because `replace` is not a valid value, which is the other open thread on this file, so take whichever ordering suits you.

helm/openmrs/templates/velero-bucket-setup.yaml
```diff
@@ -0,0 +1,59 @@
...
+            - |
+              set -e
+              echo "Creating Velero backup bucket if it does not exist..."
+              aws s3api create-bucket \
+                --bucket {{ required "backup.bucket is required when backup.enabled=true" .Values.backup.bucket }} \
+                --endpoint-url {{ .Values.backup.s3Url }} \
+                --region {{ .Values.backup.region }} \
+                2>&1 | grep -v "BucketAlreadyOwnedByYou" \
+                || { echo "Bucket already exists, continuing."; true; }
+              echo "Bucket ready."
```
dkayiwa, line 57: `set -e` on line 49 does not reach this, so the Job still reports success when the bucket was not created, which is the half of my earlier comment on `mariadb-backup-cronjob.yaml` that is still open. A pipeline's exit status is its last command's, so `grep -v` decides it, and `grep -v` succeeds exactly when the error is something other than `BucketAlreadyOwnedByYou`. The `||` branch therefore fires on the expected error and not on the unexpected ones.

I ran the rendered script with a stubbed `aws`:

- exits 255 printing `InvalidAccessKeyId`: prints the error, then "Bucket ready.", exit 0
- exits 255 printing `BucketAlreadyOwnedByYou`: "Bucket already exists, continuing.", exit 0
- succeeds: exit 0

So bad credentials or an unreachable endpoint give you a Job marked Completed with no bucket, and the next thing you hear about it is Velero's BackupStorageLocation sitting Unavailable. Branching on the exit code instead of the output gets all three right (exit 1, 0, 0 against the same stubs, and it still renders):

```suggestion
              set -e
              echo "Creating Velero backup bucket if it does not exist..."
              if OUT=$(aws s3api create-bucket \
                --bucket {{ required "backup.bucket is required when backup.enabled=true" .Values.backup.bucket }} \
                --endpoint-url {{ .Values.backup.s3Url }} \
                --region {{ .Values.backup.region }} 2>&1); then
                echo "Bucket created."
              else
                case "$OUT" in
                  *BucketAlreadyOwnedByYou*|*BucketAlreadyExists*)
                    echo "Bucket already exists, continuing." ;;
                  *)
                    echo "$OUT" >&2
                    exit 1 ;;
                esac
              fi
              echo "Bucket ready."
```

helm/openmrs/templates/velero-backup-schedule.yaml
dkayiwa: The registration PUT is there now, but it cannot succeed against the cluster this chart deploys, and the hook reports success either way.

Two independent reasons the registration fails. `path.repo` is not set anywhere in the repo, and Elasticsearch 8.15.3 refuses an `fs` repository without it: `FsRepository` throws `location [...] doesn't match any of the locations specified by path.repo because this setting is empty` (`server/src/main/java/org/elasticsearch/repositories/fs/FsRepository.java`). And even with it set, an `fs` repository has to sit on storage mounted at the same path on every master and data node; the chart's default `elasticsearch-eck.nodeSets` is 3 nodes with a per-node `ReadWriteOnce` `elasticsearch-data` claim (`kind-openmrs.yaml` drops it to 1), so on the defaults there is no shared filesystem to point it at.

The silent part is why `onError: Fail` will not save you here. `curl -s` exits 0 on an HTTP 500, so the `&&` chain runs the snapshot PUT anyway and the whole hook command exits 0. I ran the rendered command against a stub returning 500 for both PUTs and got exit 0, so Velero records the hook as successful and the Backup finishes Completed with no Elasticsearch snapshot in it.

Pointing the repository at the SeaweedFS S3 the chart already runs sidesteps both storage problems, and `repository-s3` sits under `modules/` rather than `plugins/` in the 8.15.3 tree, so it ships with the distribution and needs nothing added to the image. Registering it once at install time rather than on every backup is still the right shape, as you said originally. Whichever repository it ends up being, the curls need `--fail-with-body` or an explicit status check, or a broken hook will keep passing.

praisekavuma4:
@dkayiwa , @rkorytkowski  and @suubi-joshua have a look at my latest fix for the backup.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
```diff
@@ -0,0 +1,89 @@
...
+            - name: dump
+              emptyDir: {}
+          initContainers:
+            - name: dump
+              image: {{ .Values.backup.mariadbBackup.image | default (printf "mariadb:%s" (.Values.mariadb.image.tag | default "10.11")) }}
+              env:
+                - name: MARIADB_ROOT_PASSWORD
+                  valueFrom:
+                    secretKeyRef:
+                      name: {{ .Release.Name }}-backend-mariadb-secret
```
dkayiwa: `{{ .Release.Name }}-backend-mariadb-secret` only exists when `global.mariadb.enabled=true`, but this CronJob asks for it either way. I rendered `--set global.mariadb.enabled=false --set backup.mariadbBackup.enabled=true` and the only occurrence of that name in the whole output is this `secretKeyRef`, so on an external database the dump container would sit in CreateContainerConfigError and nothing would ever reach S3. Not a blocker, and only on that configuration.

`mariadbBackup.mariadbHost` already anticipates dumping a database the chart did not deploy, so the secret name and key look like its natural companions under `mariadbBackup`. Failing that, gating the template on `global.mariadb.enabled`, the way the other umbrella templates gate on their own infra flag, would at least make it not-render rather than not-work.

helm/openmrs/templates/velero-backup-schedule.yaml
dkayiwa: Still open on the current head: line 23 on the BackupStorageLocation and line 45 on the Schedule are both `{{ .Release.Namespace }}`. Rendering `--set backup.enabled=true --api-versions velero.io/v1/Schedule` with helm 3.17.4 puts the Secret in `velero` and those two in `openmrs`.

If merged as-is, `helm install` succeeds and neither object is ever reconciled, so no Backup is created and `kubectl get schedule -n velero` comes back empty. Nothing errors, which is the part that will cost somebody a real recovery. I re-checked the mechanism against v1.18.2: `pkg/cmd/server/server.go` builds the manager with `Cache: cache.Options{DefaultNamespaces: map[string]cache.Config{f.Namespace(): {}}}`, and the schedule controller is constructed as `NewScheduleReconciler(s.namespace, ...)` at line 882.

The Secret itself is already where it needs to be. BSL credentials are resolved through `credentials.NewNamespacedSecretStore(mgr.GetClient(), f.Namespace())` (server.go:307), so in Velero's namespace rather than the BSL's. The whole fix is `namespace: {{ .Values.backup.veleroNamespace }}` on lines 23 and 45.

The new README section needs the same edit, since it currently presents this placement as intended ("Enabling `backup.enabled=true` deploys, in the release namespace: a `BackupStorageLocation` ... a daily `Schedule`").

helm/openmrs/templates/mariadb-backup-cronjob.yaml
dkayiwa: Still here, at line 80 of the current head. Rendering with `--set backup.mariadbBackup.s3ForcePathStyle=true`, the upload container gets `PATH_STYLE_ARG="--no-sign-request"` and passes it to `aws s3 cp` alongside the credentials coming in through `envFrom`.

If merged as-is, an operator who sets `s3ForcePathStyle: true`, which is exactly what SeaweedFS and MinIO need and SeaweedFS is what this repo runs, gets an unsigned anonymous PutObject that 403s against an authenticated endpoint, while still not getting the path-style addressing the value name promises. `--no-sign-request` should just come out, and `aws configure set default.s3.addressing_style path` before the copy is what actually sets it. I re-checked botocore on that: `DEFAULT_S3_CONFIG_VARS` still declares `'addressing_style': (('s3', 'addressing_style'), None, None, None)`, the second slot being the env-var name, so it is config-file only, unlike neighbours such as `s3_disable_express_session_auth` which do carry one.

The new README row for `backup.mariadbBackup.s3ForcePathStyle` writes the flag down as documented behaviour ("adds `--no-sign-request` to the upload"), so that line needs to go with it.

helm/openmrs/templates/velero-backup-schedule.yaml
dkayiwa: Unchanged at lines 72-77 of the current head, so both halves of this still stand: an `fs` repository cannot be registered without `path.repo`, which is not set anywhere in this repo, and `curl -s` exits 0 on an HTTP 500, so the `&&` chain runs the snapshot PUT anyway and the whole hook command exits 0. `onError: Fail` never comes into play and the Backup finishes Completed with no Elasticsearch snapshot in it.

One thing has moved underneath this since I wrote it. The hook is gated on `(index .Values "openmrs-backend" "elasticsearch" "enabled")`, but after TRUNK-6736 that key is the backend's workload wiring (its ES volume, the wait loop, hibernate-search URIs, including external clusters). What actually puts Elasticsearch in the namespace is the umbrella's own `elasticsearch.enabled`, which `Chart.yaml` uses as the condition on the `eck-elasticsearch` dependency. `kind-openmrs.yaml` happens to set both, so the hook does render there, but an umbrella install with `elasticsearch.enabled: true` alone gets ES pods and no hook at all. Worth gating on the key that decides whether the pods exist.

helm/openmrs/templates/velero-crd-check.yaml
dkayiwa: Lines 17-21 and the guard they describe are unchanged, and I re-ran it on this head with helm 3.17.4: `helm template helm/openmrs --set backup.enabled=true` still fails with "Velero is not installed on this cluster" unless `--api-versions velero.io/v1/Schedule` is passed too. `.Capabilities.KubeVersion.Version` is never empty during `helm template`, so `not (empty ...)` is always true and the `else if` behaves as a plain `else`. The NOTE at the top of the file and the comment below it still say opposite things; dropping the `else if` down to `else` is the honest version, since the NOTE already covers the flag.

The CI step is the part I would most want, and this round is the argument for it. `helm-build.yaml` on main has grown a lot since I last looked and still leaves `backup.enabled` at false everywhere, so none of these five templates are rendered by anything. I ran the step from my comment above against this head: 27 resources, 22 validated, 5 skipped (the Velero CRDs have no published schema), 0 invalid. It passes as things stand, and it would have failed on the `command:` map from two rounds ago.

helm/openmrs/templates/velero-restore-cronjob.yaml
dkayiwa: The PR description is still the pre-review one. It says `Chart.yaml` was bumped 1.2.2 to 1.2.3 and gained Velero as an optional dependency, and `Chart.yaml` is not in the diff at all now; it lists `velero-restore-cronjob.yaml` as running a health check and triggering a restore only when the cluster is found to be genuinely broken, which is the behaviour that was deliberately taken out; and it does not mention `velero-bucket-setup.yaml`, `velero-crd-check.yaml` or `mariadb-backup-cronjob.yaml` at all. Worth rewriting before this merges, since it is the first thing a reviewer reads.

helm/openmrs/templates/mariadb-backup-cronjob.yaml
praisekavuma4: @dkayiwa thank you here. its been solved well. what else could be left untouched?

praisekavuma4:
@suubi-joshua , @dkayiwa  @rkorytkowski  the backup now seems ready, take a look at it.

PR #8, opened 2026-05-05
TRUNK-6535: Migrate elasticsearch from bitnami to official elastic chart

suubi-joshua:
- Switched the Elasticsearch dependency from `bitnami/elasticsearch@22.0.0` to the official `elastic/elasticsearch@8.5.1`
- Remapped all `values.yaml` parameters from Bitnami's structure to the Elastic chart's structure.
- Updated `Chart.yaml` and `Chart.lock` with the new repository (https://helm.elastic.co)
- Updated `helpers.tpl:` service name changed to `elasticsearch-master`, port path changed to .service.port
- Updated `kind-openmrs.yaml` to use replicas: 1 instead of the old master.replicaCount: 3

Jira Ticket: https://openmrs.atlassian.net/browse/TRUNK-6535


suubi-joshua:
Hi @rkorytkowski here is the PR but before i merge am still testing the ES service locally, but still hitting some local environment issues.

rkorytkowski:
Have you seen https://github.com/elastic/helm-charts#elastic-stack-kubernetes-helm-charts ? The helm you used is now deprecated. We should use eck operator as documented at https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s

suubi-joshua:
Ohh had not seen that yet, my bad let me fix that.

suubi-joshua:
In the latest commit 

- ECK operator is documented as a cluster prerequisite (not bundled as a subchart) since it installs cluster-scoped CRDs and RBAC that don't belong inside an application chart

- Plugin installation uses postStart lifecycle hook on the ES container per ECK documentation, init containers don't have access to the correct ES_HOME context

- `xpack.security.enabled: false` is opt-in via `disableSecurity` toggle (default false), only `kind-openmrs.yaml` sets it to true for local testing

Still doing local kind testing to confirm the ES pod reaches healthy state before requesting full review.

suubi-joshua:
Hi @rkorytkowski ,
I’m hitting a few roadblocks while testing the ECK Operator locally in Kind and would appreciate your guidance on two specific points:

1.  Version vs. Resources:
 Observation: `Version 8.15.3` consistently fails to start (Exit Code 1/OOM) on my local machine despite increasing heap to 512MB. Version 7.17.24 starts successfully but triggers EOL warnings.

 Should we prioritize 8.x for production parity and look into further resource stripping, or is staying on 7.17 acceptable for the local development profile and production?

2. Connectivity & Security Configuration:
- The backend pod cannot connect to Elasticsearch. I initially suspected TLS, but I have now disabled it via the Operator’s http.tls settings.
- I have removed the manual xpack.security.enabled settings from the config block to resolve a conflict where the Operator was flagging them as "reserved for internal use."
- After making these changes the Operator now reconciles successfully, but the handshake still fails. I am investigating if there is a service name mismatch or if the backend is still defaulting to HTTPS.

Relevant Logs: [https://justpaste.it/lk51l](https://www.google.com/search?q=https://justpaste.it/lk51l)

rkorytkowski:
The error log
` Response: 401 'Unauthorized' from 'http://openmrs-elasticsearch-es-http.openmrs.svc.cluster.local:9200' with `
suggests that connection is successful, but authorization fails. You either need to disable security for ES or provide hibernate.search.backend.username and hibernate.search.backend.password in openrms-core. We don't have these properties defined as `OMRS_SEARCH_...` variables, but you can pass them as `OMRS_EXTRA_...` or add support for OMRS_SEARCH_USERNAME and OMRS_SEARCH_PASSWORD to sartup-init.sh script in openmrs-core.

rkorytkowski:
I read in docs that disabling security in ECK operator is not supported. However, you can enable anonymous access instead, see https://www.elastic.co/guide/en/elasticsearch/reference/8.15/anonymous-access.html

helm/openmrs-backend/templates/elasticsearch.yaml
```diff
@@ -0,0 +1,72 @@
...
+      volumeClaimTemplates:
+        - metadata:
+            name: elasticsearch-data
+          spec:
+            accessModes:
+              - ReadWriteOnce
+            resources:
+              requests:
+                storage: {{ .Values.elasticsearch.storageSize }}
+{{- end }}
```
rkorytkowski: I wonder if using a helm chart for this wouldn't make things simpler: https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/managing-deployments-using-helm-chart#using-helm-values-to-install-only-es 

It definitely makes it easier to add additional services of ECK stack like Kibana or Logstash if needed. I would try that for this reason.

rkorytkowski:
Regarding the OOM, how much RAM do you have in your system? Are you using Docker Desktop? If yes, then make sure it doesn't have a lower memory limit than needed by containers (go to settings -> resources).

suubi-joshua:
> Regarding the OOM, how much RAM do you have in your system? Are you using Docker Desktop? If yes, then make sure it doesn't have a lower memory limit than needed by containers (go to settings -> resources).

16 GB

rkorytkowski:
How much of 16 GB is available for docker containers? Have you checked?

~~For local testing you can limit to just a single ES node.~~ I see you are already doing that.

suubi-joshua:
> How much of 16 GB is available for docker containers? Have you checked?

image

helm/openmrs-backend/values.yaml
```diff
@@ -65,77 +65,74 @@ mariadb:
...
+          - ingest
+        node.store.allow_mmap: false
+
+        # Anonymous access — ECK-supported way to allow unauthenticated
+        # connections from the OpenMRS backend without disabling security.
+        # See: https://www.elastic.co/guide/en/elasticsearch/reference/8.15/anonymous-access.html
+        # WARNING: Only enable in local/CI environments. Never use in production.
+        xpack.security.authc.anonymous.username: anonymous
+        xpack.security.authc.anonymous.roles: superuser
+        xpack.security.authc.anonymous.authz_exception: false
```
rkorytkowski: Could you please create an issue to add support for es username and password variables in startup.sh in openmrs core? Also mention in the issue to adjust helm to remove anonymous access.

rkorytkowski (approved):
Code looks good. Have you finally been able to run it locally?

rkorytkowski (approved):

helm/kind-openmrs.yaml
```diff
@@ -27,8 +27,28 @@ openmrs-backend:
...
+          xpack.security.authc.anonymous.username: anonymous
+          xpack.security.authc.anonymous.roles: superuser
+          xpack.security.authc.anonymous.authz_exception: false
+        podTemplate:
+          spec:
+            containers:
+              - name: elasticsearch
+                env:
+                  - name: ES_JAVA_OPTS
+                    value: "-Xmx512m -Xms512m"
```
rkorytkowski: Is this really needed? Isn't ES already setup to make use of `-XX` and utilise the available container memory?

helm/kind-openmrs.yaml
```diff
@@ -27,8 +27,28 @@ openmrs-backend:
...
+        count: 1
+        config:
+          node.roles:
+            - master
+            - data
+            - ingest
+          node.store.allow_mmap: false
+          xpack.security.authc.anonymous.username: anonymous
+          xpack.security.authc.anonymous.roles: superuser
+          xpack.security.authc.anonymous.authz_exception: false
```
rkorytkowski: Does it all have to be repeated in kind-openmrs?

helm/openmrs-backend/values.yaml
```diff
@@ -65,77 +65,74 @@ mariadb:
...
+              command:
+                - sh
+                - -c
+                - sysctl -w vm.max_map_count=262144
+
+          containers:
+            - name: elasticsearch
+              env:
+                - name: ES_JAVA_OPTS
+                  value: "-Xmx512m -Xms512m"
```
rkorytkowski: Is it really needed?

helm/openmrs-backend/values.yaml
```diff
@@ -65,77 +65,74 @@ mariadb:
...
+              lifecycle:
+                postStart:
+                  exec:
+                    command:
+                      - sh
+                      - -c
+                      - |
+                        if [ -n "analysis-phonetic" ]; then
+                          bin/elasticsearch-plugin install --batch analysis-phonetic || true
```
rkorytkowski: I think you can use:
```
plugins:
    - analysis-phonetic
 ```
under the nodeSets

File: helm/openmrs-backend/values.yaml [hunk 7]
```diff
@@ -65,77 +65,74 @@ mariadb:
...
+        spec:
+          initContainers:
+            - name: sysctl
+              securityContext:
+                privileged: true
+                runAsUser: 0
+              command:
+                - sh
+                - -c
+                - sysctl -w vm.max_map_count=262144
```
rkorytkowski: Is this really needed? Doesn't eck operator do it?

rkorytkowski: I don't see it being mentioned https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/elasticsearch-deployment-quickstart

rkorytkowski:
Just noticed the note about needing to assign min. of 2Gib for ES node at https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/elasticsearch-deployment-quickstart

Maybe that's the issue why it doesn't want to start for you locally.

rkorytkowski (changes requested):
Actually a few clarifications still neeeded.

helm/kind-openmrs.yaml
suubi-joshua: I had added it as a debugging measure while troubleshooting the OOM issue locally (ES 8.x was consistently OOMing and Iwas trying to bound the heap to see if that stabilised it). But you're correct that ECK uses `-XX:+`UseContainerSupport by default, so JVM already derives heap from the container memory limit. Will remove this.

helm/openmrs-backend/values.yaml
suubi-joshua: Still changes that were left in in my local debugging, am going to remove them.

helm/kind-openmrs.yaml
suubi-joshua: No, it does not. I had duplicated it in `kind-openmrs.yaml` to make the local Kind setup self-contained and easy to read in one place. But its really unnecessary because the anonymous access config is already the default in `values.yaml`.

helm/openmrs-backend/values.yaml
suubi-joshua: I wasn't aware ECK had native plugin support at the nodeSet level will change this.

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -85,18 +95,43 @@ spec:
               mountPath: /work-dir
       {{- end }}
       {{- if .Values.elasticsearch.enabled }}
+        - name: import-es-ca
```
suubi-joshua, line 49: Why I added this init container, ECK generates a self-signed CA for its Elasticsearch HTTP TLS. OpenMRS's JVM doesn't trust it by default, causing Hibernate Search to fail with PKIX path building errors. This init container imports the ECK CA certificate into the JVM truststore so the backend can connect to Elasticsearch over TLS.

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -144,6 +186,10 @@ spec:
               echo "Reading JDBC URL from /work-dir/jdbc-url.txt"
               export OMRS_DB_URL=$(cat /work-dir/jdbc-url.txt)
               echo "OMRS_DB_URL=$OMRS_DB_URL"
+              {{- if .Values.elasticsearch.enabled }}
+              export OMRS_EXTRA_HIBERNATE_SEARCH_BACKEND_PASSWORD=$(cat /etc/es-cred/elastic)
+              export JAVA_OPTS="$JAVA_OPTS -Djavax.net.ssl.trustStore=/truststore/cacerts -Djavax.net.ssl.trustStorePassword=changeit -Dhibernate.search.backend.ssl.verification_mode=none"
```
suubi-joshua: The `startup.sh` script resets `JAVA_OPTS`, so truststore and SSL  args passed via the pod command get lost. So I am setting them again here after `startup.sh` has run but before OpenMRS initializes.

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -144,6 +186,10 @@ spec:
               echo "Reading JDBC URL from /work-dir/jdbc-url.txt"
               export OMRS_DB_URL=$(cat /work-dir/jdbc-url.txt)
               echo "OMRS_DB_URL=$OMRS_DB_URL"
+              {{- if .Values.elasticsearch.enabled }}
+              export OMRS_EXTRA_HIBERNATE_SEARCH_BACKEND_PASSWORD=$(cat /etc/es-cred/elastic)
```
suubi-joshua: ECK generates a random elastic user password into a k8s secret. This reads it into `OMRS_EXTRA_*` so `startup-init.sh` converts it to `hibernate.search.backend.password` for Hibernate Search.

helm/kind-openmrs.yaml
suubi-joshua: So investigating this further, You'd expect `kind-openmrs.yaml` to only override what it specifies and inherit the rest from `values.yaml`. The `nodeSets` array from `kind-openmrs.yaml` completely replaces the `nodeSets` array from `values.yaml`. The anonymous access config from `values.yaml` is lost because the entire `nodeSets` list is swapped out, not deep-merged. So the duplication in `kind-openmrs.yaml` isn't redundant, it's the only way to ensuring that the settings survive. In my local testing If I remove `anonymous access` from `kind-openmrs.yaml`, it does not get inherited from `values.yaml` it's absent.

suubi-joshua:
All pods run succesfully and can access the web app via `http://localhost:8080`
image

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -28,9 +28,13 @@ data:
   {{- if .Values.elasticsearch.enabled }}
   OMRS_SEARCH: "elasticsearch"
   OMRS_SEARCH_ES_URIS: {{ include "openmrs-elasticsearch.url" . }}
+  # Allows Hibernate Search to accept ECK's self-signed certificate
+  OMRS_EXTRA_HIBERNATE_SEARCH_BACKEND_SSL_VERIFICATION__MODE: "none"
+  JAVA_TOOL_OPTIONS: "-Djavax.net.ssl.trustStore=/truststore/cacerts -Djavax.net.ssl.trustStorePassword=changeit"
```
rkorytkowski, line 36: Could you please explain what it does and why it is needed aside from setting "OMRS_EXTRA_HIBERNATE_SEARCH_BACKEND_SSL_VERIFICATION__MODE"?

rkorytkowski: Please add as a comment in yaml above this line.

helm/openmrs-backend/templates/statefulset.yaml
rkorytkowski: Isn't  OMRS_EXTRA_HIBERNATE_SEARCH_BACKEND_SSL_VERIFICATION__MODE: "none" working?

helm/openmrs-backend/values.yaml
rkorytkowski: Since you are passing username and password for ES to openmrs backend, anonymous access is no more needed, right?

helm/kind-openmrs.yaml
```diff
@@ -27,8 +27,41 @@ openmrs-backend:
...
+        podTemplate:
+          spec:
+            initContainers:
+              - name: install-plugins
+                command:
+                  - sh
+                  - -c
+                  - bin/elasticsearch-plugin install --batch analysis-phonetic
```
rkorytkowski: You should be using 'plugins: -analysis-phonetic' instead as in default values.

helm/kind-openmrs.yaml
rkorytkowski: What are you customizing here compared to default values?

helm/openmrs-backend/values.yaml
```diff
@@ -65,77 +65,54 @@ mariadb:
...
-        memory: 1024Mi
-    persistence:
-      enabled: true
-      size: 8Gi
+elasticsearch-eck:
+  version: "8.15.3"
+
+  nodeSets:
+    - name: default
+      count: 1
```
rkorytkowski: Probably default count should be 3.

helm/openmrs-backend/values.yaml
```diff
@@ -65,77 +65,54 @@ mariadb:
...
+
+      podTemplate:
+        spec:
+          containers:
+            - name: elasticsearch
+              resources:
+                requests:
+                  cpu: 500m
+                  memory: "2Gi"
+                  ephemeral-storage: 50Mi
```
rkorytkowski: Isn't it too low storage to start with?

rkorytkowski (changes requested):
I think some cleanup is needed. Since you pass ES username and password to backend you can remove anonymous access.

helm/openmrs-backend/values.yaml
suubi-joshua: Yes my bad will increase it to 10gi

helm/kind-openmrs.yaml
suubi-joshua: Yes this skipped my mind

helm/openmrs-backend/templates/statefulset.yaml
suubi-joshua: Yes its was not working, the PKIX errors persisted.

helm/openmrs-backend/templates/configmap.yaml
suubi-joshua: It imports the ECK CA into the JVM truststore so Hibernate Search can verify the certificate properly instead of blindly skipping verification, and ensures trust works even if verification_mode=none is removed later.

helm/openmrs-backend/values.yaml
suubi-joshua: yes

helm/kind-openmrs.yaml
suubi-joshua: Changing this in kind has actually brought up a new issue ECK operator 3.4.0's CRD does not have a `plugins:` field on `nodeSets`, the only properties are `config`, `count`, `name`, `podTemplate`, and `volumeClaimTemplates`. The Kubernetes API server silently strips unknown fields, so `plugins:` has no effect on ECK 3.x. And surely if i remove the `initContainer` the `backend` container throws Elasticsearch errors.
If we want to avoid the `initContainer:` and go with plugins I need a to do a Custom Docker image that will build a derivative image with analysis-phonetic pre-installed at build time.  Then reference it via `elasticsearch-eck.image:` in `kind-openmrs.yaml`. Zero init containers, zero runtime overhead.

rkorytkowski: Alright, let's use a pre-configured image. See https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/create-custom-images

Let's put it in https:/github.com/openmrs/openmrs-contrib-elasticsearch and publish to github docker registry with github actions.

suubi-joshua: Okay let me look into this

suubi-joshua: Hi @rkorytkowski  you can check out the PR here for this, 
https://github.com/openmrs/openmrs-contrib-elasticsearch/pull/1

helm/openmrs-backend/values.yaml
```diff
@@ -65,77 +65,46 @@ mariadb:
...
+      count: 3
+      config:
+        node.roles:
+          - master
+          - data
+          - ingest
+        node.store.allow_mmap: false
+
+      plugins:
+        - analysis-phonetic
```
rkorytkowski: I understood this is not working. Please remove it.

suubi-joshua:
I am still testing this locally with the latest changes,

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -11,13 +11,13 @@ data:
...
   {{- else if gt (int (.Values.mariadb.replicas | default 2)) 1 }}
   OMRS_DB_HOSTNAME: "{{ .Release.Name }}-mariadb-primary"
-  OMRS_DB_URL: "jdbc:mariadb:replication://{{ .Release.Name }}-mariadb-primary.{{ .Release.Namespace }}.svc.cluster.local:3306,{{ .Release.Name }}-mariadb-secondary.{{ .Release.Namespace }}.svc.cluster.local:3306/{{ .Values.mariadb.auth.database | default "openmrs" }}?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8"
+  OMRS_DB_URL: "jdbc:mariadb:replication://{{ .Release.Name }}-mariadb-primary.{{ .Release.Namespace }}.svc.cluster.local:3306,{{ .Release.Name }}-mariadb-secondary.{{ .Release.Namespace }}.svc.cluster.local:3306/{{ .Values.mariadb.auth.database | default "openmrs" }}?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true"
```
suubi-joshua, line 17: The MariaDB Connector/J v3 driver reports itself as "MariaDB" rather than "MySQL". OpenMRS Liquibase changelogs use `dbms="mysql` " [preconditions](https://github.com/search?q=org%3Aopenmrs%20%3Cdbms%20type%3D%22mysql%22%20%2F%3E&type=code), which silently skip every table-creating changeset on MariaDB. The result is a database with zero application tables. This affects any OpenMRS deployment using the MariaDB operator (not just this PR), but was never caught because the old Bitnami chart bundled an older driver that reported as MySQL. Hence this change

helm/openmrs-backend/templates/mariadb-instance.yaml
```diff
@@ -32,6 +32,7 @@ spec:
     binlog_format=ROW
     innodb_autoinc_lock_mode=2
     max_connections=200
+    max_user_connections=200
```
suubi-joshua, line 35: The MariaDB operator creates the openmrs user with a default limit of 20 connections. OpenMRS's connection pool easily exceeds this during Liquibase migrations, causing startup failures.

suubi-joshua:
The latest commited changes are not directly about Elasticsearch, they're fixes uncovered while testing the integration of the latest changes i.e with the merged MariaDB operator changes from main .
After merging main (which has the MariaDB operator from [PR #9](https://github.com/openmrs/openmrs-contrib-cluster/pull/9)) into this branch, I ran the full deployment to verify everything still works. The backend pod kept crashing during startup.

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -30,9 +30,14 @@ data:
   {{- if .Values.elasticsearch.enabled }}
   OMRS_SEARCH: "elasticsearch"
   OMRS_SEARCH_ES_URIS: {{ include "openmrs-elasticsearch.url" . }}
+  # Skips cert verification for Hibernate Search (workaround)
+  OMRS_EXTRA_HIBERNATE_SEARCH_BACKEND_SSL_VERIFICATION__MODE: "none"
```
rkorytkowski, line 34: Since you are adding a self-signed certifcate this can be removed right?

rkorytkowski (approved):
Please remove OMRS_EXTRA_HIBERNATE_SEARCH_BACKEND_SSL_VERIFICATION__MODE in some follow up commit. I'll merge this one now.

helm/openmrs-backend/values.yaml
suubi-joshua: THis should have skipped my mind.

helm/openmrs-backend/templates/configmap.yaml
suubi-joshua: Yes ideally yes but when i remove it i get hibernate errors, but will address it in the followup PR

PR #9, opened 2026-05-08
TRUNK-6520: Use MariaDB official K8s operator

suubi-joshua:
MariaDB operator migration
- Bitnami mariadb dependency removed from Chart.yaml
- MariaDB operator installed as cluster prerequisite (CRDs + operator)
- MariaDB CRD template created with Galera topology (verified minimal spec)
- Secret template created for operator-managed credentials
- ConfigMap updated with operator condition (Galera uses <release>-mariadb, replication uses <release>-mariadb-primary)
- Statefulset init container updated for operator topology
- Galera cluster formed with 3 nodes and healthy
- JDBC URL correctly points to operator service with loadbalance mode
- Liquibase migrations completed successfully

JIRA Ticket: https://openmrs.atlassian.net/browse/TRUNK-6520


suubi-joshua:
Here all pods are running health except the backend pod as well, failing with these [logs](https://justpaste.it/l59m9), OCL related, I want to investigate if the backend pod errors are a raising from java version for both this PR and the elastic migration one.  
image

README.md
```diff
@@ -53,6 +53,25 @@ Make sure that Docker is running and issue the following commands:
       kubectl -n kubernetes-dashboard port-forward svc/kubernetes-dashboard-kong-proxy 8443:443
       # Go to https://localhost:8443/ and login with generated token
 
+#### Prerequisites: MariaDB Operator
```
suubi-joshua, line 56: AM thinking of having this automated via shell rather than manual, but I dont know if its possible will look into it once these backend errors are fixed.

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         - name: workdir
           emptyDir: { }
 
-      {{ if or (.Values.galera.enabled) (.Values.mariadb.enabled) }}
+      {{ if or (.Values.galera.enabled) (.Values.mariadb.enabled) (index .Values "mariadb-operator" "enabled" | default false) }}
```
rkorytkowski: Are we still using mariadb.enabled and galera.enabled or switched to mariadb-operator.enabled?

We should drop support for mariadb.enabled and galera.enabled and remove bitnami mariadb charts.

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -48,6 +48,15 @@ spec:
             - -c
             - |
               echo "Resolving MariaDB hosts..."
+              {{ $operatorEnabled := index .Values "mariadb-operator" "enabled" | default false }}
+              {{- if $operatorEnabled }}
+              {{- $operatorGalera := index .Values "mariadb-operator" "galera" | default false }}
+              {{- if $operatorGalera }}
+              echo "jdbc:mariadb:loadbalance://{{ .Release.Name }}-mariadb.{{ .Release.Namespace }}.svc.cluster.local:3306/$OMRS_DB_NAME?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8" > /work-dir/jdbc-url.txt
+              {{- else }}
+              echo "jdbc:mariadb://{{ .Release.Name }}-mariadb-primary.{{ .Release.Namespace }}.svc.cluster.local:3306/$OMRS_DB_NAME?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8" > /work-dir/jdbc-url.txt
```
rkorytkowski: How is the replica URL passed to the backend? It is used to optimize read access by the mariadb driver, thus it was resolved below with kubectl for maridb bitnami images.

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         - name: workdir
           emptyDir: { }
 
-      {{ if or (.Values.galera.enabled) (.Values.mariadb.enabled) }}
+      {{ if or (.Values.galera.enabled) (.Values.mariadb.enabled) (index .Values "mariadb-operator" "enabled" | default false) }}
       initContainers:
         - name: resolve-mariadb-hosts
           image: bitnamilegacy/kubectl:1.33.4
```
rkorytkowski: Since we are dropping support for mariadb bitnami charts we could use kubectl from other supported image e.g. https://hub.docker.com/r/alpine/kubectl if resolution of replicas for building a URL is still needed.

helm/kind-openmrs.yaml
```diff
@@ -12,7 +12,7 @@ openmrs-backend:
   monitoring:
     enabled: true
   galera:
-    enabled: true
+    enabled: false
```
rkorytkowski: You should delete galera and mariadb bitnami config from here.

README.md
rkorytkowski: We could have a dedicated helm chart to install cluster operators named openmrs-operator.

We could use it for mariadb, elasticsearch and possibly others.

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -35,7 +47,7 @@ data:
   {{- if .Values.minio.enabled }}
   storage.type: "s3"
   storage.s3.region: "us-east-1"
-  storage.s3.endpoint: {{ include "openmrs-minio.url" . }}
+  storage.s3.endpoint: 'http://openmrs-minio.{{ .Release.Namespace }}.svc.cluster.local:9000'
```
rkorytkowski: Why are you changing this? It doesn't seem to be related.

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -5,6 +5,17 @@ metadata:
   labels:
     {{- include "openmrs-backend.labels" . | nindent 4 }}
 data:
+  {{- $operatorEnabled := index .Values "mariadb-operator" "enabled" | default false }}
```
rkorytkowski: Can we just call it mariadb or does it have to be mariadb-operator?

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -5,6 +5,17 @@ metadata:
...
+  {{- if $operatorEnabled }}
+  OMRS_DB: "mariadb"
+  OMRS_DB_NAME: {{ .Values.mariadb.auth.database | default "openmrs" | quote }}
+  {{- if $operatorGalera }}
+  OMRS_DB_HOSTNAME: "{{ .Release.Name }}-mariadb"
+  {{- else }}
+  OMRS_DB_HOSTNAME: "{{ .Release.Name }}-mariadb-primary"
+  {{- end }}
+  {{- else }}
   {{- if and (not .Values.mariadb.enabled) (not .Values.galera.enabled) }}
```
rkorytkowski: These are no longer supported, right?

rkorytkowski (changes requested):
Going in the good direction, but need to get rid of old bitnami charts settings and maybe rename mariadb-operator setting to be just mariadb.

helm/openmrs-backend/templates/configmap.yaml
suubi-joshua: Yes its not my minio pod was failing to start because the way the port number was accessed. Thats why i made the change, but will revert it. Its not meant to be part of this PR

helm/openmrs-backend/templates/configmap.yaml
suubi-joshua: Yes

README.md
suubi-joshua: Okay, I think i will do this in a different PR or can add it here though I feel if i add it here it might delay this from being merged.

rkorytkowski: Sounds good to do it in a separate issue and PR

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -48,36 +48,11 @@ spec:
...
-                MODE="loadbalance"
-              {{ else }}
-                MODE="replication"
-              {{ end }}
-              echo "jdbc:mariadb:$MODE://$HOSTS:3306/$OMRS_DB_NAME?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8" > /work-dir/jdbc-url.txt
+              {{- if .Values.mariadb.galera }}
+              echo "jdbc:mariadb:loadbalance://{{ .Release.Name }}-mariadb.{{ .Release.Namespace }}.svc.cluster.local:3306/$OMRS_DB_NAME?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8" > /work-dir/jdbc-url.txt
+              {{- else }}
+              echo "jdbc:mariadb:replication://{{ .Release.Name }}-mariadb-primary.{{ .Release.Namespace }}.svc.cluster.local:3306,{{ .Release.Name }}-mariadb-secondary.{{ .Release.Namespace }}.svc.cluster.local:3306/$OMRS_DB_NAME?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8" > /work-dir/jdbc-url.txt
+              {{- end }}
```
rkorytkowski: It seems we don't need this init container. The URL is not constructed at runtime based on available pods. It can be passed directly to openmrs-backend.

rkorytkowski (changes requested):

rkorytkowski:
Almost there. I think we just need to get rid of the init container.

helm/openmrs-backend/templates/statefulset.yaml
rkorytkowski: Will you be fixing this?

rkorytkowski (changes requested):

helm/openmrs-backend/templates/statefulset.yaml
suubi-joshua: Yes

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -5,26 +5,25 @@ metadata:
...
+  {{- if .Values.mariadb.enabled }}
+  OMRS_DB: "mariadb"
+  OMRS_DB_NAME: {{ .Values.mariadb.auth.database | default "openmrs" | quote }}
+  OMRS_DB_PORT: "3306"
+  {{- if .Values.mariadb.galera }}
+  OMRS_DB_HOSTNAME: "{{ .Release.Name }}-mariadb"
+  OMRS_DB_URL: "jdbc:mariadb:loadbalance://{{ .Release.Name }}-mariadb.{{ .Release.Namespace }}.svc.cluster.local:3306/{{ .Values.mariadb.auth.database | default "openmrs" }}?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8"
+  {{- else }}
+  OMRS_DB_HOSTNAME: "{{ .Release.Name }}-mariadb-primary"
+  OMRS_DB_URL: "jdbc:mariadb:replication://{{ .Release.Name }}-mariadb-primary.{{ .Release.Namespace }}.svc.cluster.local:3306,{{ .Release.Name }}-mariadb-secondary.{{ .Release.Namespace }}.svc.cluster.local:3306/{{ .Values.mariadb.auth.database | default "openmrs" }}?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8"
```
rkorytkowski, line 17: Do we always have a secondary replica? What happens if we use replicas: 1?

rkorytkowski (changes requested):

helm/openmrs-backend/templates/configmap.yaml
suubi-joshua: Nice catch No. We don't always have a secondary replica.
Currently the replicas are hardcoded to 2 in non-Galera mode, so replicas: 1 isn't possible right now. But if someone did set it to 1, the JDBC URL would still try connecting to `<name>-mariadb-secondary` which wouldn't exist, and the backend would fail to start.

I'll make replicas configurable and only include the secondary in the JDBC URL when replicas > 1.

PR #10, opened 2026-05-11
fix: Resolve backend pod crash, 404 on localhost, and helm template bugs

suubi-joshua:
### 1. Localhost 404: Grafana ingress shadowing OpenMRS
When accessing `http://localhost:8080/openmrs/`, nginx returned **404 Not Found**. The Grafana ingress had `host: localhost`, which created an nginx `server_name localhost` block that shadowed the OpenMRS catch-all ingress (`server_name _;`). Requests for `/openmrs/` hit the Grafana block which only knows `/grafana`, producing a 404.

**Fix:** Changed Grafana's ingress host to `grafana.local` (matching its `server.domain`). The OpenMRS catch-all now correctly handles `localhost` requests.

### 2. Backend pod crash loop: Hibernate Search connection failure
The backend pod would start, Hibernate Search would immediately try connecting to Elasticsearch at `http://openmrs-elasticsearch:9200`, but the ES cluster (3 master replicas) was still forming quorum. Spring's context refresh failed with a `BeanCreationException`, the webapp undeployed, and the pod entered a crash/restart loop.

**Fix:** Added a `wait-for-elasticsearch` initContainer that uses `kubectl get statefulset` to block pod startup until the ES master statefulset reports ready replicas (up to 5 min timeout). This ensures ES is available before OpenMRS initializes Hibernate Search.

### 3. ConfigMap rendered literal text instead of value
When using external ES (`elasticsearch.enabled: false, elasticsearch.uris: ...`), the ConfigMap contained the literal string `.Values.elasticsearch.uris` instead of the actual URI value, because it was written as a raw YAML string instead of a Helm template expression.

**Fix:** Wrapped the value with `{{ ... | quote }}` so Helm evaluates it properly.

### 4. Malformed ES/MinIO URLs: misplaced `quote` function
The `_helpers.tpl` URL templates used `quote .Values.elasticsearch.service.ports.restAPI`, which wraps the port number in literal double-quote characters. The rendered URL became `http://service.namespace.svc.cluster.local:"9200"` — syntactically invalid.

**Fix:** Replaced `quote ...` with `(.Values... | toString)` to produce a clean string without embedded quotes.

### 5. Typo in kind config
`defaultIngessClass` was missing an 'r', which would cause it to be ignored by nginx.

**Fix:** Corrected to `defaultIngressClass`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Prepend with the name of the service: `openmrs-backend`, `openmrs-frontend`, `op
 | `openmrs-backend.monitoring.enabled`                             | Enable monitoring (Grafana, Loki, Alloy)                                                                               | `"false"`                                                 |
 | `openmrs-backend.grafana.adminPassword`                          | Grafana admin password                                                                                                 | `"Admin123"`                                              |
 | `openmrs-backend.grafana.ingress.enabled`                        | Enable ingress for Grafana                                                                                             | `"true"`                                                  |
-| `openmrs-backend.grafana.ingress.hosts`                          | Hosts for Grafana ingress                                                                                              | `["localhost"]`                                           |
+| `openmrs-backend.grafana.ingress.hosts`                          | Hosts for Grafana ingress                                                                                              | `["grafana.local"]`                                       |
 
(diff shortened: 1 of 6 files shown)


suubi-joshua:
These were blocking my local testing, are they valid or is there an easier work around

PR #12, opened 2026-05-20
TRUNK 6484: Migrate edge routing from Nginx Ingress to Traefik Gateway API

suubi-joshua:
## **Summary**

This PR replaces the legacy `ingress-nginx` controller with the Kubernetes Gateway API using Traefik v3.x. All `kind: Ingress `resources have been removed in favor of native HTTPRoute resources, and the deployment model is restructured for proper namespace isolation.

## What Changed
New chart: `helm/openmrs-gateway/`
- Declares Traefik Helm chart v34 as a dependency
- Configures Gateway API provider, disables legacy Ingress provider
- Creates a Gateway resource (`openmrs-gateway`) in `kube-system` with `allowedRoutes.namespaces.from: All`
- Maps listener `port 8000` to match Traefik's web entry point

Backend & Frontend Routing (`helm/openmrs-backend/`, `helm/openmrs-frontend/`)
- Deleted `templates/ingress.yaml` from both charts, no more `kind: Ingress`
- Added `templates/httproute.yaml`, native Gateway API resources
- Backend: `PathPrefix /openmrs/` → service port 8080
- Frontend: `PathPrefix /openmrs/spa` → URLRewrite (ReplacePrefixMatch /) → service port 80
- Removed all nginx annotations, `ingressClassName`, and `global.defaultIngressClass`

Kind (`helm/kind-init.yaml`, `kind-openmrs.yaml`)
- Removed embedded `ingress-nginx` manifest from `kind-init.yaml`
- Added `openmrs-gateway.traefik.ports.web.hostPort: 80` for local Kind access
- No `defaultIngressClass` references remain

CRD Bootstrap
- Switched to `experimental-install.yaml` for Gateway API CRDs (required for URLRewrite filter support)

- Enforced mandatory two-layer `install: gateway` first in kube-system, then apps in openmrs namespace
- Updated README with correct install order and port-forward instructions

- Gateway listener port: changed from `80` to `8000` to match Traefik's web entry point (`port 80` has no matching entry point)
- `HTTPRoute backendRefs[].port:` corrected from nested object `port: { number: X }` to plain integer `port: X`
- Backend readiness probe: changed from /openmrs/health/started to /openmrs/health/alive, the "started" endpoint was returning 503 until OCL import completes, preventing the backend from receiving traffic even though the app serves requests correctly

JIRA Ticket: https://openmrs.atlassian.net/browse/TRUNK-6484


helm/openmrs-backend/values.yaml
```diff
@@ -240,7 +229,7 @@ livenessProbe:
   periodSeconds: 10
 readinessProbe:
   httpGet:
-    path: /openmrs/health/started
+    path: /openmrs/health/alive
```
rkorytkowski: Actually, we only want the user to go to the service once any initialisation is done and the user can log in... Please use the /started endpoint.

helm/openmrs-frontend/templates/httproute.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if .Values.ingress.enabled }}
```
rkorytkowski: Rename to gateway.enabled

helm/openmrs-frontend/values.yaml
```diff
@@ -48,16 +45,7 @@ service:
 
 ingress:
```
rkorytkowski: Use "gateway:" for gateway and leave ingress as is, but disable it by default

helm/openmrs-backend/templates/httproute.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.ingress.enabled }}
```
rkorytkowski: Use gateway.enabled for gateway and leave ingress for ingress.

helm/openmrs-backend/templates/httproute.yaml
```diff
@@ -0,0 +1,21 @@
...
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ include "openmrs-backend.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs-backend.labels" . | nindent 4 }}
+spec:
+  parentRefs:
+    - name: openmrs-gateway
```
rkorytkowski: I feel that it should be named traefik-gateway. There's nothing specific to openmrs in the gateway setup. Same as we had it for ingress. Typically you just install one traefik for your cluster to serve all services.

helm/openmrs-backend/templates/ingress.yaml
```diff
@@ -1,65 +0,0 @@
-{{- if .Values.ingress.enabled -}}
```
rkorytkowski, line 1: Do not delete it. We shouldn't be removing support for ingress in our templates. Some K8s providers may still expose services via ingress. We just want to use traefik gateway instead of nginx ingress in default setup.

helm/openmrs-backend/values.yaml
```diff
@@ -4,7 +4,6 @@
 
 global:
   defaultStorageClass: "gp2"
-  defaultIngressClass: "nginx"
```
rkorytkowski, line 7: Shouldn't we specify defaultGatewayClass: "traefik"?

helm/openmrs-backend/values.yaml
```diff
@@ -207,17 +206,7 @@ service:
 
 ingress:
   enabled: true
```
rkorytkowski, line 166: Let's not remove support for ingress in charts. Just set enabled: false.

helm/openmrs-frontend/templates/ingress.yaml
```diff
@@ -1,65 +0,0 @@
-{{- if .Values.ingress.enabled -}}
```
rkorytkowski, line 1: Again let's not remove ingress support. It will be disabled by default.

helm/openmrs-gateway/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: openmrs-gateway
```
rkorytkowski: I'd rather rename the helm to traefik-gateway. There's nothing specific to OpenMRS in traefik gateway and it can be used across your whole cluster. We provide the helm setup as a convenience in case K8s provider doesn't do it.

rkorytkowski (changes requested):

helm/openmrs-backend/values.yaml
suubi-joshua: Okay makes sense now.

rkorytkowski:
Do you run into any issues with this? Why is it still draft?

helm/openmrs-backend/templates/httproute.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: HTTPRoute
+metadata:
+  name: {{ include "openmrs-backend.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs-backend.labels" . | nindent 4 }}
+spec:
+  parentRefs:
+    - name: traefik-gateway
+      namespace: kube-system
```
rkorytkowski: These two should be configurable via properties. The gateway may be installed elsewhere.

helm/openmrs-frontend/templates/httproute.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: HTTPRoute
+metadata:
+  name: {{ include "openmrs-frontend.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openmrs-frontend.labels" . | nindent 4 }}
+spec:
+  parentRefs:
+    - name: traefik-gateway
+      namespace: kube-system
```
rkorytkowski: These two should be configurable via props.

helm/openmrs-frontend/values.yaml
```diff
@@ -3,7 +3,7 @@
 # Declare variables to be passed into your templates.
 
 global:
-  defaultIngressClass: "nginx"
+  defaultIngressClass: "traefik"
```
rkorytkowski: I don't see defaultIngressClass and defaultGatewayClass being used anywhere. If that's the case, let's get rid of them. defaultIngressClass was a property used by bitnami helm charts. If we don't have usage for this property then let's drop it.

rkorytkowski (changes requested):

suubi-joshua:
> Do you run into any issues with this? Why is it still draft?

Yes the pods run well but I dont seem to be going past the login page, SO am still trying to investigate why. I have some few commits locally will push them today and will shift it from Draft.

helm/openmrs-frontend/values.yaml
suubi-joshua: `defaultIngressClass` was orphaned when I initially deleted the `ingress` templates. But i had restored the ingress templates and `defaultIngressClass` is used again at `ingress.yaml` to set `ingressClassName`.

suubi-joshua:
These are the [backend logs](https://justpaste.it/jdt7o), the queue module is failing to start.

rkorytkowski:
@suubi-joshua the error doesn't seem to be related to traefik rather DB issues. Is it failing after changes from this PR or before as well?

suubi-joshua:
> @suubi-joshua the error doesn't seem to be related to traefik rather DB issues. Is it failing after changes from this PR or before as well?

Yes it is not related, just that its new it was not happening before. But seems not related to the work here, i just try to make sure that everything works well before any this is merged. And when i tested on main everything was working well.

But I have tested again and things seem to be working well, the error seems to have disappeared. But I cant seem to be going past login page.

suubi-joshua:
@rkorytkowski 
Need your guidance on something,
Login fails with 2+ backend replicas because CSRFGuard stores tokens in-memory per pod. Traefik round-robins requests, so the login POST hits a different pod than the one that issued the CSRF token, causing validation failure.

I have tried adding `traefik.ingress.kubernetes.io/service.sticky.cookie: "true"` annotation on the backend HTTPRoute. This had no effect, the Gateway API provider does not honour this annotation.

Locally replacing the backend HTTPRoute with an `IngressRoute` + `TraefikService` (Traefik CRDs) fixes this issue. The `TraefikService` configures cookie‑based stickiness (`openmrs_session`) which is fully supported by the CRD provider. The frontend HTTPRoute for `/openmrs/spa` stays unchanged, it doesn't need stickiness.

We could also do a shared session store but that would require changing the OpenMRS Docker image (Tomcat session replication config) and adding Redis or a session DB, which is far more invasive. The routing fix is a zero‑change deploy‑side solution that works at any replica count.

The Traefik's Gateway API provider does not reliably support `sticky.cookie` even via `TraefikService` backendRefs.

Should i go ahead with replacing the backend HTTPRoute with an `IngressRoute` + `TraefikService` (Traefik CRDs)

suubi-joshua:
In the latest commit I Replaced the HTTPRoute with a TraefikService that configures cookie-based stickiness (openmrs_session) routed via an IngressRoute.

rkorytkowski:
Have you tried following official docs at https://doc.traefik.io/traefik/expose/kubernetes/advanced/#using-gateway-api-with-traefikservice ? I didn't get the chance to inspect your code for sticky sessions before changing to ingress. The docs I linked provide an example of sticky sessions config with gateway api so it should be supported. Ingress may go away at some point, but it's still supported so it's fine to use it if we can't make the gateway work.

suubi-joshua:
> Have you tried following official docs at https://doc.traefik.io/traefik/expose/kubernetes/advanced/#using-gateway-api-with-traefikservice ? I didn't get the chance to inspect your code for sticky sessions before changing to ingress. The docs I linked provide an example of sticky sessions config with gateway api so it should be supported. Ingress may go away at some point, but it's still supported so it's fine to use it if we can't make the gateway work.

I got it working. The backend HTTPRoute now references a TraefikService which handles cookie-based stickiness. Tested with 2 backend replicas: requests without the cookie are load-balanced across pods, requests with the cookie consistently hit the same pod (same` JSESSIONID`). The frontend HTTPRoute stays as-is with URLRewrite for `/openmrs/spa`.

I also made the sticky cookie's secure flag configurable via `traefikService.sticky.cookie.secure` (defaults to false for dev, should be set true in production behind HTTPS).

helm/openmrs-backend/templates/httproute.yaml
```diff
@@ -0,0 +1,22 @@
...
+      namespace: {{ .Values.gateway.parentRefs.namespace }}
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: {{ .Values.gateway.path | default "/openmrs/" }}
+      backendRefs:
+        - group: traefik.io
+          kind: TraefikService
+          name: {{ include "openmrs-backend.fullname" . }}-sticky
```
rkorytkowski, line 22: Since this backendRef is traefik specific it should also be added with {{- if .Values.traefikService.enabled }}.

rkorytkowski (changes requested):
Small change needed.

helm/openmrs-backend/templates/httproute.yaml
suubi-joshua: I have opted to guard the entire `HTTPRoute` with `and .Values.gateway.enabled .Values.traefikService.enabled` instead of just the` backendRef.` . If `traefikService` is disabled and the HTTPRoute falls back to a regular Service, the backend would lose sticky sessions, which breaks login with 2+ replicas.  I figured that guarding the whole route avoids creating a broken-by-default state.

suubi-joshua:
LGTM everything works just well, i will be making a PR for SeaweedFS to replace MiniIO and then openmrs operator  to automate prerequisites.

helm/openmrs-backend/templates/httproute.yaml
rkorytkowski: Does it work if traefikService is disabled and gateway enabled? I would think that this route fails in such a case, thus suggested to put "if .Values.traefikService.enabled" around TraefikService backendRef and add else with openmrs service backendRef. If you don't want to run multiple replicas of backend then you can opt for not using TraefikService.

suubi-joshua: Yes the route would fail in such a senario
@rkorytkowski I went with guarding the entire HTTPRoute rather than just the backendRef. My reasoning is if traefikService is disabled and the HTTPRoute falls back to a regular Service, a user with 2+ backend replicas would lose sticky sessions, causing login failures (CSRF token mismatch across pods). Guarding the whole route avoids creating a route that silently breaks login.

That said, I see your point, single-replica deployments or users with shared session storage don't need stickiness, and the fallback keeps Gateway API available without TraefikService. 
Do you prefer the targeted route with the else fallback, or is the global guard acceptable?

rkorytkowski: Let me ask it differently. What happens if gateway is enabled, traefikSerice is disabled and backend replicas is set to 1 with your latest changes? Is there still a gateway api in front of the backend service with a single replica in such a case? If it works, then I'm happy with the solution. I'm just not seeing the HTTPRoute to backend service in such a scenario, but I may be misreading the config...

suubi-joshua: You are absolutely right I missed this. With the current global guard, setting `traefikService.enabled: false` with 1 replica orphans the backend from Gateway API entirely, there's no HTTPRoute at all. I'll switch to the targeted approach you originally suggested.

suubi-joshua: I made the required change just made one twick, traefikService is automatically activated for 2+ backend pods, also automatic activation of sticky session for 2+ backends.

rkorytkowski (approved):

PR #14, opened 2026-06-09
TRUNK: 6650 Consolidate into one-command make deploy with local registry

suubi-joshua:
This PR consolidates the local Kind development flow into a single `make deploy` command that bootstraps a full OpenMRS stack idempotently. Previously, setting up a local cluster required manually running Kind, installing operators, loading images, and deploying OpenMRS in separate steps with no shared script,  the flow was scattered across ad-hoc commands and outdated helper files.

The core change is a new helm/Makefile with `deploy`, `deploy-operators`, `deploy-openmrs`, and `teardown` targets, backed by helm/scripts/bootstrap.sh (the full pipeline) and helm/scripts/lib.sh (shared helpers). The operator dependencies (MariaDB operator, ECK operator, Traefik) are bundled into a new helm/openmrs-operator/ umbrella chart so they deploy as one unit.
A key infrastructure decision was replacing kind load docker-image with a local Docker registry (kind-registry:5000). The `kind-config.yaml` was updated with `containerdConfigPatches` to configure the registry mirror.

JIRA TICKET: https://openmrs.atlassian.net/browse/TRUNK-6650

Cleaned out dead code: the old traefik-gateway/ chart (superseded by the official Traefik chart in openmrs-operator/), kind-init.yaml (local-path-provisioner and K8s Dashboard manifests, both ship built-in with Kind v0.27+), kind-openmrs-min.yaml, kind-traefik.yaml, kind-traefik-override.yaml, and helm/openmrs/helm_build.sh.


suubi-joshua:
After running `make deploy` on a clean Kind cluster, openmrs-backend-0 hangs because of three issues compound: the wait-`for-mariadb init` container used `nc -z` on port 3306 and returned success before InnoDB was ready (~12s gap), Liquibase then deadlocked on index DDL during warm-up due to innodb_lock_wait_timeout=50 being too low for Galera, and the readiness probe (`failureThreshold: 1`) killed the pod before the 5-8 minute migration could complete, all triggered by the mutable `mariadb:10.11` tag pulling a newer patch with different InnoDB timing. 
The [fix](https://github.com/openmrs/openmrs-contrib-cluster/pull/14/commits/883255a5129075cfbf22c11921f6503272545a78) replaces the TCP-only init container with credential-verified SELECT 1 using `mariadb:11.4 (--ssl=0)`, adds `innodb_lock_wait_timeout=120` to the MariaDB CR, relaxes the readiness probe (`failureThreshold 10`, `initialDelaySeconds: 60`), and pins the MariaDB image to a digest. After the fix, 1690 changesets run with zero deadlocks, the pod reaches 1/1 Ready in ~6 minutes, and the health endpoint returns 200

helm/openmrs-backend/templates/statefulset.yaml
```diff
@@ -44,8 +44,40 @@ spec:
...
+                i=$((i + 1))
+                if [ $i -ge $MAX_RETRIES ]; then
+                  echo "Timeout waiting for MariaDB after $((MAX_RETRIES * RETRY_INTERVAL)) seconds."
+                  exit 1
+                fi
+                echo "  MariaDB not ready yet (attempt $i/$MAX_RETRIES) — retrying in ${RETRY_INTERVAL}s..."
+                sleep $RETRY_INTERVAL
+              done
+              echo "MariaDB is ready and application credentials are valid."
+      {{- end }}
```
rkorytkowski: I'd like to have this script in openmrs-core startup.sh instead. The application should properly test connection before starting up without additional scripts.

rkorytkowski (changes requested):
Please fix openmrs-core db readiness check instead of initcontainer.

helm/openmrs-backend/templates/statefulset.yaml
suubi-joshua: Opened PR https://github.com/openmrs/openmrs-core/pull/6193

rkorytkowski (commented):
Looks good to me! I'll merge once I test it out locally.

suubi-joshua:
I see the SeaweedFS PR has gone in, I also need to update this PR.

PR #15, opened 2026-06-12
TRUNK:6632 Add helm template rendering validation to helm-build workflow

suubi-joshua:
Add three helm template rendering checks after packaging to prevent rendering regressions from going undetected until deployment:

- Default values (all optional features off)
- Full-featured (monitoring + Elasticsearch + MinIO enabled)
- Minimal (all optional features explicitly disabled)

Also add the elastic Helm repo and update subchart dependencies before template validation to ensure all deps (minio OCI, grafana,loki, alloy, ECK) are available for rendering.

JIRA TICKET: https://openmrs.atlassian.net/browse/TRUNK-6632

### DIFF
--- a/.github/workflows/helm-build.yaml
+++ b/.github/workflows/helm-build.yaml
@@ -23,11 +23,18 @@ jobs:
       - name: Add Helm Repositories
         run: |
           helm repo add grafana https://grafana.github.io/helm-charts
+          helm repo add elastic https://helm.elastic.co
+          helm repo update
 
       - name: Lint Subcharts
         run: |
           helm lint helm/openmrs-backend
           helm lint helm/openmrs-frontend
+
+      - name: Update Subchart Dependencies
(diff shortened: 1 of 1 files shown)


rkorytkowski (approved):

suubi-joshua:
This validation step exposes a `fromToml` function incompatibility with Helm 3.14.2 — it was added in Helm 3.17.0 (Jan 2025). The SeaweedFS subchart 4.31.0 uses `fromToml` in its security config template, which fails at parse time with older Helm versions.

We need to bump the Helm version in this workflow from `v3.14.2` to `v3.17.0` (or later). There are no breaking changes between 3.14 and 3.17. Helm maintains strong backward compatibility within the 3.x line.

Alternatively, we could switch to a version range like `v3.17` so CI always gets the latest 3.17.x patch.

PR #19, opened 2026-06-24
TRUNK-6652: Add openmrs-tenant chart for multi-tenancy

suubi-joshua:
## Summary
Adds `helm/openmrs-tenant`, a chart that deploys an isolated OpenMRS tenant (backend + frontend) sharing the primary cluster's infrastructure.

Following [#28](https://github.com/openmrs/openmrs-contrib-cluster/pull/28): which made openmrs-backend / openmrs-frontend infra-free, reusable charts, this is a thin umbrella: it depends on those two charts (2.0.0) and supplies only the per-tenant configuration. It does not fork or re-declare any backend/frontend templates, so tenants inherit the reviewed shared-chart behaviour with no drift.

## What a tenant provides
`openmrs-backend.db.url` / `.username` / `.password` — connection to the shared MariaDB (per-tenant database + user)
`openmrs-frontend.spaPath` / `.apiUrl` / `.defaultLocale` / `.configUrls` — SPA config `app.kubernetes.io`/tenant pod label via podLabels
Each tenant is a separate Helm release in its own namespace.

JIRA Ticket: https://openmrs.atlassian.net/browse/TRUNK-6652

### DIFF
--- a/.github/workflows/helm-build.yaml
+++ b/.github/workflows/helm-build.yaml
@@ -104,3 +104,15 @@ jobs:
           fail_if_renders openmrs-backend.seaweedfs.s3.enableAuth=true
           fail_if_renders openmrs-backend.seaweedfs.admin.ingress.enabled=true
           fail_if_renders openmrs-backend.seaweedfs.admin.enabled=true
+
+      - name: Lint and Render Tenant Chart
+        run: |
+          TENANT_VALUES="--set tenant.name=ci \
+            --set backend.db.hostname=mariadb.example.svc.cluster.local \
+            --set backend.db.database=openmrs_ci \
+            --set backend.db.user=ci_user \
+            --set backend.db.password=ci-pass"
+          helm lint helm/openmrs-tenant $TENANT_VALUES
+          helm template helm/openmrs-tenant --generate-name $TENANT_VALUES
+          helm template helm/openmrs-tenant --generate-name $TENANT_VALUES \
+            --set backend.infinispan.clustered=true
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ separate ConfigMaps, Secrets, Services, and labelled pods.
 #### Prerequisites
 
 - Primary OpenMRS stack deployed and running (steps 1–4 above)
-- MariaDB accessible from tenant namespace (default DNS: `openmrs-mariadb.openmrs.svc.cluster.local`)
+- MariaDB accessible from tenant namespace (default DNS: `<primary-release>-mariadb.<primary-namespace>.svc.cluster.local`, e.g. `openmrs-mariadb.openmrs.svc.cluster.local`)
 - A database and user created for the tenant:
 
 ```bash
@@ -126,7 +126,7 @@ helm install <tenant> helm/openmrs-tenant \
   -n tenant-<tenant> --create-namespace \
   --set tenant.name=<tenant> \
   --set global.defaultStorageClass=standard \
-  --set backend.db.hostname=openmrs-mariadb.openmrs.svc.cluster.local \
+  --set backend.db.hostname=<primary-release>-mariadb.<primary-namespace>.svc.cluster.local \
   --set backend.db.database=openmrs_<tenant> \
   --set backend.db.user=<tenant>_user \
   --set backend.db.password=<password>
@@ -183,11 +183,20 @@ kubectl port-forward -n tenant-<tenant> svc/<tenant>-frontend 8081:80
 | `backend.db.database` | Database name | **required** |
 | `backend.db.user` | Database user | **required** |
 | `backend.db.password` | Database password | **required** |
-| `backend.storage.size` | PVC size for /openmrs/data | `8Gi` |
-| `backend.storage.storageClass` | Backend PVC StorageClass (defaults to `global.defaultStorageClass`) | `""` |
+| `backend.persistence.size` | PVC size for /openmrs/data | `8Gi` |
+| `backend.persistence.storageClass` | Backend PVC StorageClass (defaults to `global.defaultStorageClass`) | `""` |
+| `backend.storage.type` | Storage type for patient documents | `"local"` |
+| `frontend.apiUrl` | Frontend API URL for SPA backend calls | `"http://localhost:8080/openmrs"` |
 | `frontend.image.repository` | Frontend image | `openmrs/openmrs-reference-application-3-frontend` |
 | `frontend.image.tag` | Frontend image tag | `nightly-core-2.8` |
 | `frontend.replicaCount` | Frontend Deployment replicas | `1` |
+
+> **Note on `storage.type: "local"`:** When `backend.storage.type` is `"local"` (the default),
+> each StatefulSet pod mounts its own dedicated PVC. If `backend.replicaCount > 1`,
+> patient documents uploaded to pod-0 will not be visible on pod-1. For multi-replica
+> setups, configure external shared storage (e.g. S3-compatible SeaweedFS) by setting
+> `backend.storage.type` and the corresponding S3 credentials. SeaweedFS integration
+> is currently available in the parent `openmrs` chart; tenant chart support is planned.
 
 ### Alternative: install from Helm registry
 
--- a/README.md
+++ b/README.md
@@ -145,6 +145,11 @@ helm install coast helm/openmrs-tenant \
   --set backend.db.password=CoastPass123
 ```
 
+> **Note:** Resource names (`<tenant>-backend`, `<tenant>-frontend`) derive from the
+> **release name** — the first argument to `helm install` — not from `tenant.name`.
+> `tenant.name` only sets the `app.kubernetes.io/tenant` label. Keep the release name
+> and `tenant.name` the same (as shown above) so the commands below resolve correctly.
+
 #### Verification
 
 ```bash
@@ -162,8 +167,12 @@ kubectl port-forward -n tenant-<tenant> svc/<tenant>-frontend 8081:80
 ```
 
 > **Note on routing:** Currently each tenant's backend and frontend are separate
-> services in their own namespace. You access them via `kubectl port-forward`
-> as shown above. In the future, tenant-specific Gateway API HTTPRoute resources
+> services in their own namespace. The backend API is reachable via `kubectl
+> port-forward` as shown above. The **frontend SPA UI is not fully usable over
+> port-forward yet** — the app shell requests its assets under `SPA_PATH`
+> (`/openmrs/spa`), which needs a gateway rewrite (stripping the prefix before
+> nginx) that port-forward cannot provide. In the future, tenant-specific
+> Gateway API HTTPRoute resources
 > will be added to place the frontend and backend behind the same hostname
 > (e.g. `<tenant>.example.com`) with path-based routing — `/openmrs/spa` to the
 > frontend and `/openmrs` to the backend — eliminating the need for
@@ -173,7 +182,7 @@ kubectl port-forward -n tenant-<tenant> svc/<tenant>-frontend 8081:80
 
 | Name | Description | Default |
 |------|-------------|---------|
-| `tenant.name` | Tenant identifier (used as resource prefix) | **required** |
+| `tenant.name` | Tenant identifier (sets the `app.kubernetes.io/tenant` label on every resource) | **required** |
 | `global.defaultStorageClass` | Default StorageClass for PVCs | `""` |
 | `backend.image.repository` | Backend image | `openmrs/openmrs-reference-application-3-backend` |
 | `backend.image.tag` | Backend image tag | `nightly-core-2.8` |
@@ -187,6 +196,9 @@ kubectl port-forward -n tenant-<tenant> svc/<tenant>-frontend 8081:80
 | `backend.persistence.storageClass` | Backend PVC StorageClass (defaults to `global.defaultStorageClass`) | `""` |
 | `backend.storage.type` | Storage type for patient documents | `"local"` |
 | `frontend.apiUrl` | Frontend API URL for SPA backend calls | `"http://localhost:8080/openmrs"` |
+| `frontend.spaPath` | URL path the SPA is served from (`SPA_PATH`) | `"/openmrs/spa"` |
+| `frontend.defaultLocale` | Default UI locale (`SPA_DEFAULT_LOCALE`) | `"en"` |
+| `frontend.configUrls` | Distro config JSON URLs (`SPA_CONFIG_URLS`); omitted when empty | `""` |
 | `frontend.image.repository` | Frontend image | `openmrs/openmrs-reference-application-3-frontend` |
 | `frontend.image.tag` | Frontend image tag | `nightly-core-2.8` |
 | `frontend.replicaCount` | Frontend Deployment replicas | `1` |
--- a/README.md
+++ b/README.md
@@ -98,6 +98,97 @@ To disable monitoring (Grafana, Loki, Alloy), set `monitoring.enabled=false` in 
 | `SKIP_OPERATORS=true` | `false` | Skip `openmrs-operator` chart install |
 | `SKIP_OPENMRS=true` | `false` | Skip OpenMRS deployment (exit after step 3) |
 
+### 6. Deploy additional tenants (multi-tenancy)
+
+The `helm/openmrs-tenant` chart deploys an isolated OpenMRS tenant (backend + frontend)
+that shares the primary cluster's MariaDB. Each tenant gets its own namespace with
+separate ConfigMaps, Secrets, Services, and labelled pods.
+
+#### Prerequisites
+
+- Primary OpenMRS stack deployed and running (steps 1–4 above)
+- MariaDB accessible from tenant namespace (default DNS: `openmrs-mariadb.openmrs.svc.cluster.local`)
+- A database and user created for the tenant:
+
+```bash
+kubectl exec -n openmrs svc/openmrs-mariadb -- mysql -u root -pRoot123 -e "
+  CREATE DATABASE IF NOT EXISTS openmrs_<tenant> CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
+  CREATE USER IF NOT EXISTS '<tenant>_user'@'%' IDENTIFIED BY '<password>';
+  GRANT ALL PRIVILEGES ON openmrs_<tenant>.* TO '<tenant>_user'@'%';
+  FLUSH PRIVILEGES;
+"
+```
+
+#### Install a tenant
+
+```bash
+helm install <tenant> helm/openmrs-tenant \
+  -n tenant-<tenant> --create-namespace \
+  --set tenant.name=<tenant> \
+  --set global.defaultStorageClass=standard \
+  --set backend.db.hostname=openmrs-mariadb.openmrs.svc.cluster.local \
+  --set backend.db.database=openmrs_<tenant> \
+  --set backend.db.user=<tenant>_user \
+  --set backend.db.password=<password>
+```
+
+Example for a tenant named `coast`:
+
+```bash
+helm install coast helm/openmrs-tenant \
+  -n tenant-coast --create-namespace \
+  --set tenant.name=coast \
+  --set global.defaultStorageClass=standard \
+  --set backend.db.hostname=openmrs-mariadb.openmrs.svc.cluster.local \
+  --set backend.db.database=openmrs_coast \
+  --set backend.db.user=coast_user \
+  --set backend.db.password=CoastPass123
+```
+
+#### Verification
+
+```bash
+# Check pods with tenant label
+kubectl get pods -n tenant-<tenant> -L app.kubernetes.io/tenant
+
+# Wait for ready
+kubectl wait --for=condition=ready pod -n tenant-<tenant> --all --timeout=600s
+
+# Port-forward to backend (API)
+kubectl port-forward -n tenant-<tenant> svc/<tenant>-backend 8080:8080
+
+# Port-forward to frontend (SPA)
+kubectl port-forward -n tenant-<tenant> svc/<tenant>-frontend 8081:80
+```
+
+> **Note on routing:** Currently each tenant's backend and frontend are separate
+> services in their own namespace. You access them via `kubectl port-forward`
+> as shown above. In the future, tenant-specific Gateway API HTTPRoute resources
+> will be added to place the frontend and backend behind the same hostname
+> (e.g. `<tenant>.example.com`) with path-based routing — `/openmrs/spa` to the
+> frontend and `/openmrs` to the backend — eliminating the need for
+> port-forwarding, just like the primary OpenMRS stack.
+
+#### Tenant chart parameters
+
+| Name | Description | Default |
+|------|-------------|---------|
+| `tenant.name` | Tenant identifier (used as resource prefix) | **required** |
+| `global.defaultStorageClass` | Default StorageClass for PVCs | `""` |
+| `backend.image.repository` | Backend image | `openmrs/openmrs-reference-application-3-backend` |
+| `backend.image.tag` | Backend image tag | `nightly-core-2.8` |
+| `backend.replicaCount` | Backend StatefulSet replicas | `1` |
+| `backend.db.hostname` | MariaDB hostname | **required** |
+| `backend.db.port` | MariaDB port | `3306` |
+| `backend.db.database` | Database name | **required** |
+| `backend.db.user` | Database user | **required** |
+| `backend.db.password` | Database password | **required** |
+| `backend.storage.size` | PVC size for /openmrs/data | `8Gi` |
+| `backend.storage.storageClass` | Backend PVC StorageClass (defaults to `global.defaultStorageClass`) | `""` |
+| `frontend.image.repository` | Frontend image | `openmrs/openmrs-reference-application-3-frontend` |
+| `frontend.image.tag` | Frontend image tag | `nightly-core-2.8` |
+| `frontend.replicaCount` | Frontend Deployment replicas | `1` |
+
 ### Alternative: install from Helm registry
 
       helm repo add openmrs https://openmrs.github.io/openmrs-contrib-cluster/
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ kubectl port-forward -n tenant-<tenant> svc/<tenant>-frontend 8081:80
 | `tenant.name` | Tenant identifier (sets the `app.kubernetes.io/tenant` label on every resource) | **required** |
 | `global.defaultStorageClass` | Default StorageClass for PVCs | `""` |
 | `backend.image.repository` | Backend image | `openmrs/openmrs-reference-application-3-backend` |
-| `backend.image.tag` | Backend image tag | `nightly-core-2.8` |
+| `backend.image.tag` | Backend image tag | `3.7.x-no-demo` |
 | `backend.replicaCount` | Backend StatefulSet replicas | `1` |
 | `backend.db.hostname` | MariaDB hostname | **required** |
 | `backend.db.port` | MariaDB port | `3306` |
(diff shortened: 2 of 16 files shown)


helm/openmrs-tenant/templates/frontend-service.yaml
```diff
@@ -0,0 +1,15 @@
...
+    {{- include "openmrs-tenant.frontend.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.frontend.service.type }}
+  ports:
+    - port: {{ .Values.frontend.service.port }}
+      targetPort: http
+      protocol: TCP
+      name: http
+  selector:
+    {{- include "openmrs-tenant.frontend.selectorLabels" . | nindent 4 }}
```
rkorytkowski, line 15: Did you just copy the templates from regular deployment or do I need to carefully review them?

rkorytkowski (changes requested):
Please point me to what is not copied from regular deployment and needs a careful review.

helm/openmrs-tenant/templates/_helpers.tpl
```diff
@@ -0,0 +1,73 @@
...
+{{- end }}
+
+{{- define "openmrs-tenant.backend.labels" -}}
+helm.sh/chart: {{ include "openmrs-tenant.chart" . }}
+{{ include "openmrs-tenant.backend.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/tenant: {{ required "tenant.name is required" .Values.tenant.name }}
```
suubi-joshua, line 29: All resources carry `app.kubernetes.io/tenant: {{ .Values.tenant.name }}` for filtering across namespaces.

helm/openmrs-tenant/templates/backend-configmap.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "openmrs-tenant.backend.fullname" . }}
+  labels:
+    {{- include "openmrs-tenant.backend.labels" . | nindent 4 }}
+data:
+  OMRS_DB: "mariadb"
+  OMRS_DB_HOSTNAME: {{ required "backend.db.hostname is required" .Values.backend.db.hostname | quote }}
```
suubi-joshua, line 9: Always renders individual `OMRS_DB_*` fields (hostname, port, database, user). The parent chart conditionally drops these inside a `db.url` branch; the tenant always renders them because the DB URL is always external.

helm/openmrs-tenant/templates/backend-secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "openmrs-tenant.backend.fullname" . }}
+  labels:
+    {{- include "openmrs-tenant.backend.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  OMRS_DB_USERNAME: {{ required "backend.db.user is required" .Values.backend.db.user | quote }}
```
suubi-joshua, line 9: I used Helm `required` on all DB credentials (`backend.db.hostname`, `.database`, `.user`, `.password`), so `helm install` fails immediately if any are omitted (vs. parent which defaults to internal MariaDB).

helm/openmrs-tenant/templates/frontend-service.yaml
suubi-joshua: I've left inline comments on the files with the key differences, but in summary: most templates follow the parent charts' patterns; what's new is no bundled MariaDB/ES/SF/monitoring (tenant relies on shared infrastructure), always-rendered DB fields in the configmap, `required` validation on DB credentials, per-component `storageClass` override in values.yaml, tenant labels on all resources, and empty DB defaults to prevent accidental cross-tenant credential sharing.

suubi-joshua:
@rkorytkowski  one very important change from parent charts is unlike the parent chart, there's no MariaDB, ES, SeaweedFS, or monitoring subchart. The tenant relies entirely on the primary cluster's shared services.

helm/openmrs-tenant/templates/backend-statefulset.yaml
```diff
@@ -0,0 +1,91 @@
...
+          resources:
+            {{- toYaml .Values.backend.resources | nindent 12 }}
+          envFrom:
+            - configMapRef:
+                name: {{ include "openmrs-tenant.backend.fullname" . }}
+            - secretRef:
+                name: {{ include "openmrs-tenant.backend.fullname" . }}
+          volumeMounts:
+            - name: data
+              mountPath: /openmrs/data
```
rkorytkowski, line 70: **Bug: unconditional `volumeMount` with a conditional `volumeClaimTemplates` — pod is rejected when `persistence.enabled: false`.**

The `data` volumeMount at line 66–68 is always rendered, but the `volumeClaimTemplates` block that creates the `data` volume is gated at line 81 on `{{- if .Values.backend.persistence.enabled }}`. There is no `spec.template.spec.volumes` stanza providing an `emptyDir` fallback.

When `persistence.enabled: false`, Kubernetes admission rejects the StatefulSet: `spec.containers[0].volumeMounts[0].name: Not found: "data"`. The pod never starts.

Fix — add an `emptyDir` fallback in the pod spec when persistence is off:
```yaml
      {{- if not .Values.backend.persistence.enabled }}
      volumes:
        - name: data
          emptyDir: {}
      {{- end }}
```

helm/openmrs-tenant/templates/frontend-configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "openmrs-tenant.frontend.fullname" . }}
+  labels:
+    {{- include "openmrs-tenant.frontend.labels" . | nindent 4 }}
+data:
+  SPA_PATH: "/openmrs/spa"
+  API_URL: "/openmrs"
```
rkorytkowski: **Bug: relative `API_URL` is non-functional with the documented port-forward access method.**

The README instructs users to access the frontend via:
```
kubectl port-forward svc/<tenant>-frontend 8081:80
```
With `API_URL: "/openmrs"`, the browser at `localhost:8081` resolves `/openmrs` against the same origin — hitting the frontend nginx on port 8081, not the backend on port 8080. The tenant chart has no HTTPRoutes, no Ingress, and no nginx proxy config to forward these requests. Every SPA API call returns 404 from nginx.

The README notes that HTTPRoutes "will be added in the future." Until then, the documented workflow is non-functional as shipped.

The minimum fix for the port-forward use case is to document that users must also override `API_URL` to an absolute URL:
```bash
helm install coast helm/openmrs-tenant \
  --set backend.db.hostname=... \
  --set frontend.apiUrl=http://localhost:8080
```
Or better: add a `frontend.apiUrl` value that defaults to an absolute URL, and expose it in the ConfigMap, so the chart is functional out of the box.

helm/openmrs-tenant/values.yaml
```diff
@@ -0,0 +1,149 @@
+tenant:
+  name: ""
+  hostname: ""
+
+global:
+  defaultStorageClass: "gp2"
```
rkorytkowski: **Bug: `global.defaultStorageClass: "gp2"` is AWS-specific — causes PVC to stay `Pending` on kind, GKE, AKS, and bare-metal.**

`gp2` is an AWS EBS StorageClass. It does not exist on kind (this repo's own CI cluster uses `kind-config.yaml` with the `standard` local-path provisioner). The sibling charts (`openmrs-backend`, `openmrs`) also default to `gp2`, but they are shielded by `kind-openmrs.yaml` which overrides `global.defaultStorageClass: standard`. **No equivalent `kind-tenant.yaml` override file exists for this chart**, so any kind deployment will produce a `Pending` PVC and a StatefulSet pod that never schedules.

Additionally, because `helm/openmrs-tenant/templates/backend-statefulset.yaml` line 87 always emits a quoted `storageClassName` (even when empty), there is no path to inherit the cluster's default StorageClass — the PVC always carries an explicit class name.

Fix options:
1. Change the default to `""` and add a `{{- if .Values.backend.persistence.storageClass }}` guard so the `storageClassName` field is omitted when empty (letting Kubernetes use the cluster default).
2. Add a `helm/kind-tenant.yaml` overlay that overrides `global.defaultStorageClass: standard`, mirroring the pattern in `kind-openmrs.yaml`.
3. At minimum, call out the required override in the README install command.

rkorytkowski (commented):
This PR introduces a new `helm/openmrs-tenant/` chart for lightweight multi-tenant deployments sharing the primary MariaDB. Three confirmed bugs — one blocks the chart from working when persistence is disabled, one renders the frontend non-functional under the documented access method, and one causes PVC provisioning to silently fail on every non-AWS cluster including kind.

rkorytkowski:
## Code Review

This PR adds a new `openmrs-tenant` Helm chart that deploys an isolated OpenMRS backend + frontend per tenant, sharing the primary cluster's MariaDB while omitting ES, SeaweedFS, and monitoring. The structure is clean and follows the existing chart conventions closely, but the chart has several correctness bugs that would make an out-of-the-box install non-functional.

---

### Findings

**`README.md:188-189` — Tenant parameter table documents wrong value paths for PVC configuration**

The table lists `backend.storage.size` and `backend.storage.storageClass`, but the actual keys in `values.yaml` are `backend.persistence.size` and `backend.persistence.storageClass`. These are different trees — `--set backend.storage.size=20Gi` is silently accepted by Helm and ignored; the PVC is created at the default 8 Gi with no error.

---

**`helm/openmrs-tenant/values.yaml:97` — `frontend.apiUrl: "/openmrs"` default makes the frontend non-functional via port-forward**

The README instructs `kubectl port-forward svc/coast-frontend 8081:80`. When a user opens `localhost:8081`, the SPA's API calls go to `localhost:8081/openmrs` — the frontend nginx, not the backend. Without a gateway routing `/openmrs` to the backend (explicitly deferred to future work), the entire frontend is broken for data access. The value needs to be set to `http://localhost:8080/openmrs` for the port-forward use case, or to an in-cluster service URL like `http://coast-backend:8080/openmrs` for cluster-internal use. Neither is the default.

---

**`helm/openmrs-tenant/templates/backend-configmap.yaml:22` — `storage.type: "local"` is hardcoded with no corresponding key in `values.yaml`**

There is no `backend.storage.type` value — the literal string `"local"` is embedded in the template with no override path. Setting `replicaCount: 2` makes this worse: each StatefulSet pod gets its own PVC via `volumeClaimTemplates`, so uploaded patient documents on pod-0 are invisible to pod-1. If SeaweedFS integration is intentionally out of scope, the README description should be updated; if it is in scope, a `backend.storage.type` value and the corresponding S3 credential path need to be added.

---

**`README.md:112` — MariaDB DNS name in prerequisites assumes the primary chart is installed with release name `openmrs`**

The documented hostname `openmrs-mariadb.openmrs.svc.cluster.local` only resolves when the primary chart's release name is `openmrs`. The MariaDB service is named `{{ .Release.Name }}-mariadb`, so a release named `openmrs-backend` produces `openmrs-backend-mariadb.openmrs.svc.cluster.local`. The prerequisite and example `--set backend.db.hostname=...` command should show the hostname pattern, not a hardcoded value.

---

**`helm/openmrs-tenant/templates/` (missing file) — No `frontend-serviceaccount.yaml` template exists, but `values.yaml` exposes `frontend.serviceAccount.create`**

`_helpers.tpl` defines `openmrs-tenant.frontend.serviceAccountName` with a full `create: true` branch, and `frontend-deployment.yaml` uses that helper for `spec.serviceAccountName`. When `create: true`, the helper returns a computed name (e.g. `coast-frontend`), but no template ever creates that ServiceAccount — the Deployment references a resource that doesn't exist, leaving all frontend pods in `CreateContainerConfigError`. The backend has the full `backend-serviceaccount.yaml` (SA + Role + RoleBinding); the frontend has nothing.

---

**`helm/openmrs-tenant/templates/backend-serviceaccount.yaml:14` — RBAC `statefulset-reader` Role and RoleBinding are created for every install regardless of `infinispan.clustered`**

The Role granting `get/list/watch` on `statefulsets` exists solely to support JGroups KUBE_PING peer discovery, which is only needed when `infinispan.clustered: true`. As written, every tenant install — including the simple `replicaCount: 1, clustered: false` default — grants the backend SA API access to StatefulSet resources unnecessarily. The RBAC block should be wrapped in `{{- if .Values.backend.infinispan.clustered }}`.

suubi-joshua:
@rkorytkowski Quick clarification on the `statefulset-reader` RBAC,

You are suggesting gating the Role/RoleBinding on `infinispan.clustered`, but looking at the parent `openmrs-backend` chart, it creates them unconditionally. The parent's approach seems deliberate, JGroups DNS_PING doesn't actually need this RBAC, so it's harmless when clustering is off.
I've implemented the gate, but want to confirm, would this be an intentional improvement over the parent, or should the tenant match the parent's unconditional pattern?

If yes, should I apply the same fix to the parent openmrs-backend chart for consistency?

rkorytkowski:
> @rkorytkowski Quick clarification on the `statefulset-reader` RBAC,
> 
> You are suggesting gating the Role/RoleBinding on `infinispan.clustered`, but looking at the parent `openmrs-backend` chart, it creates them unconditionally. The parent's approach seems deliberate, JGroups DNS_PING doesn't actually need this RBAC, so it's harmless when clustering is off. I've implemented the gate, but want to confirm, would this be an intentional improvement over the parent, or should the tenant match the parent's unconditional pattern?
> 
> If yes, should I apply the same fix to the parent openmrs-backend chart for consistency?

@suubi-joshua you are right, it's not for infinispan. Claude was wrong. It's for example for Kubernetes dashboard or ES. Drop the gate.

helm/openmrs-tenant/templates/backend-configmap.yaml
```diff
@@ -0,0 +1,22 @@
...
+  {{- if .Values.backend.db.url }}
+  OMRS_DB_URL: {{ .Values.backend.db.url | quote }}
+  {{- else }}
+  OMRS_DB_URL: {{ printf "jdbc:mariadb://%s:%s/%s?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true" .Values.backend.db.hostname (toString .Values.backend.db.port) .Values.backend.db.database | quote }}
+  {{- end }}
+  {{- if .Values.backend.infinispan.clustered }}
+  cache.type: "cluster"
+  {{- else }}
+  cache.type: "local"
+  {{- end }}
```
dkayiwa, line 21: Is clustering actually meant to work in the tenant chart? This branch sets `cache.type: cluster` but drops the three keys the parent's configmap sets right next to it: `cache.stack: kubernetes`, `jgroups.dns.query`, and `jgroups.bind.port`. The statefulset and service still open the infinispan ports, and the backend SA still gets statefulset read access, but with no `jgroups.dns.query` the pods have nothing to discover each other by. So at `backend.replicaCount > 1` they never form one cache cluster; each pod comes up as its own isolated cache, and nothing errors out to signal it.

Since the chart is built around a single replica per tenant, I'd suggest dropping the infinispan plumbing (this branch, plus the container/service ports and the values block) rather than shipping a half-wired feature. If you'd rather keep clustering as a future option, it needs those parent keys ported over before it does anything. Not a blocker either way.

Same shape with `backend.autoscaling.enabled` (and the frontend one), for what it's worth: there's no HPA template in the chart, so enabling it just removes the `replicas` field and pins you to one pod without creating an autoscaler. If the chart stays single-replica, that knob can go too.

dkayiwa (commented):
Nothing here blocks merge; the chart templates cleanly for the single-tenant flow the README documents (given the pre-created database it lists as a prerequisite).

The one concern that has no line in this diff to sit on: the new chart isn't wired into the helm workflows. `helm-build.yaml` lints and renders the `openmrs` umbrella chart (and lints `openmrs-backend`/`openmrs-frontend`), but it never touches `openmrs-tenant`, and since this is a standalone chart rather than a subchart of `openmrs`, nothing in CI actually renders it. So the green "Helm Build" check on this PR hasn't exercised this chart at all, and a later edit that breaks templating here would pass CI unnoticed. `helm-release.yaml` is similar: its version-bump `sed` lists the other four charts but not `openmrs-tenant`, so once chart-releaser publishes it, it stays pinned at 0.1.0 while the rest advance. Adding a `helm lint` + `helm template ... --set` step (with the required DB values) for the tenant chart, and adding it to the release `sed`, would close both. Not blocking, but cheap to do while the chart is new.

helm/openmrs-tenant/templates/backend-statefulset.yaml
```diff
@@ -0,0 +1,98 @@
...
+          securityContext:
+            {{- toYaml .Values.backend.securityContext | nindent 12 }}
+          image: "{{ .Values.backend.image.repository }}:{{ .Values.backend.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.backend.image.pullPolicy }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.backend.service.port }}
+              protocol: TCP
+            {{- if .Values.backend.infinispan.clustered }}
+            - name: infinispan-hibernate
```
dkayiwa: This can't be applied as written: `infinispan-hibernate` is 20 characters and Kubernetes caps port names at 15, so any tenant deployed with `backend.infinispan.clustered=true` gets rejected at apply time. Rendering the chart with that flag and applying it against a 1.34 API server gives:

    StatefulSet.apps "coast-backend" is invalid: spec.template.spec.containers[0].ports[1].name: Invalid value: "infinispan-hibernate": must be no more than 15 characters
    Service "coast-backend" is invalid: spec.ports[1].targetPort: Invalid value: "infinispan-hibernate": must be no more than 15 characters

The `targetPort` on line 16 of `backend-service.yaml` trips the same limit. Worth knowing why this looks safe in `openmrs-backend`: there the long name only ever lands on a Service *port name*, where the cap is 63, and the `targetPort` is numeric. Container port names and string `targetPort`s are the 15-character kind, and this chart introduced both.

Dropping the infinispan plumbing, as in the other thread, is what I'd do and it settles this too. If you keep it, the names have to fit, e.g. `ispn-hibernate`.

helm/openmrs-tenant/templates/frontend-configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "openmrs-tenant.frontend.fullname" . }}
+  labels:
+    {{- include "openmrs-tenant.frontend.labels" . | nindent 4 }}
+data:
+  SPA_PATH: "/openmrs/spa"
+  API_URL: "{{ .Values.frontend.apiUrl }}"
```
dkayiwa: Was leaving out `SPA_CONFIG_URLS` and `SPA_DEFAULT_LOCALE` deliberate? `openmrs-frontend`'s configmap sets both, and the image does ship `config-core_demo.json` at the nginx root, so a tenant frontend comes up with no distro config where the primary stack loads the RefApp one.

The locale is the part I'd double check: `startup.sh` falls back to `en_GB` when `SPA_DEFAULT_LOCALE` is unset, so tenants quietly default to a different locale than everything else in this repo, which I doubt anyone chose. Not pushing the demo config at real tenants seems sensible, but if that's the intent it would be more useful as a value tenants can point at their own config with.

README.md
```diff
@@ -98,6 +98,106 @@ To disable monitoring (Grafana, Loki, Alloy), set `openmrs-backend.monitoring.en
...
+> will be added to place the frontend and backend behind the same hostname
+> (e.g. `<tenant>.example.com`) with path-based routing — `/openmrs/spa` to the
+> frontend and `/openmrs` to the backend — eliminating the need for
+> port-forwarding, just like the primary OpenMRS stack.
+
+#### Tenant chart parameters
+
+| Name | Description | Default |
+|------|-------------|---------|
+| `tenant.name` | Tenant identifier (used as resource prefix) | **required** |
```
dkayiwa: `tenant.name` isn't the resource prefix, the release name is. With `helm install coast ... --set tenant.name=other`, the objects still render as `coast-backend` and `coast-frontend`, and only the `app.kubernetes.io/tenant` label carries `other`. That matters a little, because the verification commands above use `svc/<tenant>-backend`, which resolves only while the release name and the tenant name happen to match, and nothing in the chart or the docs requires that they do. Nothing to hold the PR for, but the row should say what the value actually does:

```suggestion
| `tenant.name` | Tenant identifier (sets the `app.kubernetes.io/tenant` label on every resource) | **required** |
```

If you'd rather have one knob than two that can drift, deriving the label from `.Release.Name` and dropping `tenant.name` would do it.

helm/openmrs-tenant/templates/frontend-configmap.yaml
dkayiwa: The absolute `apiUrl` doesn't get the port-forward flow working, and the reason sits upstream of the API calls. With `SPA_PATH` at `/openmrs/spa`, the app shell's `index.html` asks for `/openmrs/spa/openmrs.<hash>.js`, `/openmrs/spa/importmap.json`, `/openmrs/spa/routes.registry.json` and so on, but there's no `openmrs/` directory in the image at all: those 246 files sit at the nginx root. Running the image with exactly this ConfigMap's env:

    curl -o /dev/null -w '%{http_code}' localhost/openmrs/spa/openmrs.5e8008de18eb92c3.js    404
    curl -o /dev/null -w '%{http_code}' localhost/openmrs.5e8008de18eb92c3.js                200

What bridges that in the primary stack is the `URLRewrite` / `ReplacePrefixMatch: /` filter on `openmrs-frontend`'s HTTPRoute, which strips `/openmrs/spa` before nginx sees the request. A `kubectl port-forward` has nothing in the path to do that, so following the README you get a 200 on `/` (which is also why the readiness probe is satisfied and the `kubectl wait` step passes) and then a blank page. So the planned tenant HTTPRoute needs that rewrite filter, not just the route.

Adding those routes now is the option I'd pick. It also lets `apiUrl` go back to the relative `/openmrs` and puts the SPA on the same origin as the API, which matters because openmrs-core's `web.xml` registers no CORS filter, so a cross-origin `apiUrl` is the next thing that would break. If routing has to stay out of scope for this PR, then `SPA_PATH` needs to become a value instead of being hardcoded here, and the README plus NOTES.txt should stop offering port-forward as a way to reach the SPA.

helm/openmrs-tenant/templates/backend-configmap.yaml
suubi-joshua: You're right the clustered branch drops `cache.stack` / `jgroups.dns.query` / `jgroups.bind.port`, so at `replicaCount > 1` it wouldn't form one cache. Full clustering (discovery + sticky sessions) and the HPA templates i plan to handle under this [ticket](https://openmrs.atlassian.net/browse/TRUNK-6655) in this chart's roadmap. For this PR I am keeping the `infinispan` / `autoscaling` value keys as inert stubs (both default off, no runtime effect) to avoid a breaking values in the future when i actually handle this issue, and fixed only the port-name bug so a `clustered=true` render is valid. Worth noting: [#22](https://github.com/openmrs/openmrs-contrib-cluster/pull/22) trys to fix the parent's clustering, it was broken, so what I intend to do is while working on  [ticket](https://openmrs.atlassian.net/browse/TRUNK-6655) I will wire tenant clustering using that PR's headless-service pattern rather than the old parent config.

helm/openmrs-tenant/templates/backend-statefulset.yaml
suubi-joshua: Nice catch @dkayiwa I am setting container ports to `ispn-hibernate` (14) / `ispn-api` (8), under the 15-char limit. I also switched the Service `targetPorts` to numeric (7800/7801), which lets the Service port names stay as `infinispan-hibernate` / `infinispan-api`, matching `openmrs-backend`, which is exactly why the parent never hit this (long names only ever land on Service port names, where the cap is 63, with numeric `targetPorts`).

helm/openmrs-tenant/templates/frontend-configmap.yaml
suubi-joshua: I have added  `SPA_DEFAULT_LOCALE` (new `frontend.defaultLocale` value, default "en" to match the primary stack rather than the image's silent `en_GB` fallback) and `SPA_CONFIG_URLS` (new `frontend.configUrls` value). `configUrls` defaults to empty and is omitted from the `ConfigMap` when unset, so no demo config is pushed at real tenants, each tenant points it at its own trusted config. Also made `SPA_PATH` a value while I was there.

With `SPA_PATH=/openmrs/spa` and nothing stripping the prefix, the app shell `404`s and you get a blank page even with an absolute `apiUrl`. The full rewrite for this is when I work on this [ticket](https://openmrs.atlassian.net/browse/TRUNK-6654). For now I've made SPA_PATH a value and updated the README routing note to be explicit that port-forward reaches the backend API only, the SPA UI isn't fully usable until I handle https://openmrs.atlassian.net/browse/TRUNK-6654. I am doing this to not have a massive PR, I want to handle small chunks at a time.

helm/openmrs-tenant/templates/frontend-configmap.yaml
suubi-joshua: `frontend.apiUrl` now defaults to an absolute URL. The full port-forward SPA flow needs one more thing though, the app shell requests assets under `SPA_PATH (/openmrs/spa)`, which needs a gateway rewrite stripping the prefix before nginx, and port-forward can't do that. I'm handling the routing (HTTPRoute + ReplacePrefixMatch) under [TRUNK-6654](https://openmrs.atlassian.net/browse/TRUNK-6654) to keep this PR small. For now I've made SPA_PATH a value and updated the README routing note to say port-forward reaches the backend API only.

.github/workflows/helm-build.yaml
```diff
@@ -58,4 +58,16 @@ jobs:
...
+        run: |
+          TENANT_VALUES="--set tenant.name=ci \
+            --set backend.db.hostname=mariadb.example.svc.cluster.local \
+            --set backend.db.database=openmrs_ci \
+            --set backend.db.user=ci_user \
+            --set backend.db.password=ci-pass"
+          helm lint helm/openmrs-tenant $TENANT_VALUES
+          helm template helm/openmrs-tenant --generate-name $TENANT_VALUES
+          helm template helm/openmrs-tenant --generate-name $TENANT_VALUES \
+            --set backend.infinispan.clustered=true
```
dkayiwa, line 82: Worth knowing that these three commands pass even when the clustered manifests are invalid. Against the pre-fix chart at 8ed625f, where the container port was still named `infinispan-hibernate`, `helm lint` and both `helm template` calls all exit 0. The 15-character cap lives in the API server's Go validation rather than in the published schema (the OpenAPI definitions for `ContainerPort.name` and `ServicePort.targetPort` carry no length constraint), so it only shows up on apply:

```
StatefulSet.apps "coast-backend" is invalid: spec.template.spec.containers[0].ports[1].name:
  Invalid value: "infinispan-hibernate": must be no more than 15 characters
```

which also rules out a schema validator as the cheap alternative: kubeconform calls all 10 resources of that same broken render valid. `persistence.enabled: false` is the same shape, since the missing `data` volume was an admission rejection too, so adding it to the render list wouldn't buy much.

Covering this needs a real API server. The repo already carries `helm/kind-config.yaml`, so a throwaway kind cluster plus `helm template ... | kubectl apply --dry-run=server -f -` over the same value sets would do it. I'd keep that as a follow-up rather than grow this PR; the step as written still earns its place for template errors. Nothing here blocks merge.

helm/openmrs-tenant/templates/frontend-configmap.yaml
dkayiwa: The README caveat didn't make it into NOTES.txt, and NOTES.txt is the copy the operator actually reads, since Helm prints it the moment the install succeeds. Rendering the chart today still ends with:

```
To port-forward for local access:
  kubectl -n tenant-coast port-forward svc/coast-backend 8080:8080
  kubectl -n tenant-coast port-forward svc/coast-frontend 8081:80
```

so that second line still presents itself as a working way into the UI, and whoever follows it gets the blank page instead of the explanation now in the README. A clause on that line, saying the SPA assets 404 until the HTTPRoute work under TRUNK-6654 lands, would close it. Not a blocker.

suubi-joshua: Nice one, I wonder how i forgot this, thanks @dkayiwa

PR #20, opened 2026-07-01
TRUNK-6671: Expose Grafana, SeaweedFS dashboards and root-to-SPA redirect via Gateway API HTTPRoutes.

suubi-joshua:
Adds Gateway API HTTPRoute templates to expose Grafana, SeaweedFS Master, and SeaweedFS Filer dashboards via Traefik, plus a root `localhost:8080/` redirect to the OpenMRS SPA, all accessible out of the box on `localhost:8080`.

  
 JIRA TICKET: https://openmrs.atlassian.net/browse/TRUNK-6671
 
 
[Screencast from 03-07-26 11:29:05.webm](https://github.com/user-attachments/assets/82ca5546-9fac-4c27-8835-6b5602cc028d)


helm/openmrs-backend/values.yaml
```diff
@@ -256,8 +273,18 @@ grafana:
   ingress:
     enabled: true
```
rkorytkowski: probably should be disabled by default to match other resources since we are adding httpRoute.

helm/openmrs-backend/values.yaml
```diff
@@ -256,8 +273,18 @@ grafana:
...
     hosts:
-      - "grafana.local"
+      - "localhost"
     path: '/grafana'
+  route: {}
+  # Custom HTTPRoute config used by grafana-httproute.yaml
+  httpRoute:
+    enabled: false
```
rkorytkowski, line 277: Do we have httpRoute disabled for other resources by default? We need to be consistent.

README.md
```diff
@@ -66,7 +66,17 @@ Once deployment completes, OpenMRS is available at:
...
       http://localhost:8080/openmrs/spa/login
 
-No port-forwarding needed — Traefik binds the port directly.
+With monitoring enabled (default in `kind-openmrs.yaml`), additional dashboards are available at:
+
+| Service | URL |
+|---------|-----|
+| Grafana | http://localhost:8080/grafana/ |
+| SeaweedFS Master | http://localhost:8080/seaweedfs-master/ |
+| SeaweedFS Filer | http://localhost:8080/seaweedfs-filer/ |
```
rkorytkowski: Is filer providing a web UI or needed by the UI?
Is seaweedfs-master providing a web UI?

helm/openmrs-backend/values.yaml
suubi-joshua: Yes my bad.

helm/openmrs-backend/values.yaml
suubi-joshua: All 4 routes already default to `enabled: false`. The naming variance (httpRoute vs route vs no wrapper) is justified by Grafana subchart owns the route key, so I am using `httpRoute`  to avoid collision. SeaweedFS has no such conflict, so route is fine i think. BUt for consistency let me rename them to `httpRoute`

README.md
suubi-joshua: Both serve web UIs. Master provides cluster topology/volume visibility useful for ops. Filer provides a file browser (partial static assets break without /seaweedfsstatic/ route but am trying to fix this).

suubi-joshua: After reading the docs, I think a better approach is to use weed admin (the unified administrative dashboard) the same way we use Grafana, with a native `-urlPrefix` flag, instead of routing individual master and filer UIs. I have enabled this in the latest commit.

helm/openmrs-frontend/templates/root-redirect-httproute.yaml
```diff
@@ -0,0 +1,30 @@
...
+            type: PathPrefix
+            value: /
+      filters:
+        - type: RequestRedirect
+          requestRedirect:
+            scheme: http
+            statusCode: 302
+            path:
+              type: ReplacePrefixMatch
+              replacePrefixMatch: {{ .Values.rootRedirect.targetPath }}
```
rkorytkowski: **Bug: `ReplacePrefixMatch` on `PathPrefix: /` produces concatenated paths without a separator.**

Gateway API's `ReplacePrefixMatch` removes the matched prefix and prepends the replacement with whatever remains. For prefix `/`, only the leading slash is consumed, so `/favicon.ico` → `favicon.ico` → `/openmrs/spa/home` + `favicon.ico` = `/openmrs/spa/homefavicon.ico`. Only the exact path `/` redirects correctly.

Fix: change the match to `type: Exact, value: /` and the filter to `type: ReplaceFullPath, replaceFullPath: /openmrs/spa/home`.

```yaml
matches:
  - path:
      type: Exact
      value: /
filters:
  - type: RequestRedirect
    requestRedirect:
      statusCode: 302
      path:
        type: ReplaceFullPath
        replaceFullPath: /openmrs/spa/home
```

helm/openmrs-frontend/templates/root-redirect-httproute.yaml
```diff
@@ -0,0 +1,30 @@
...
+  {{- end }}
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /
+      filters:
+        - type: RequestRedirect
+          requestRedirect:
+            scheme: http
```
rkorytkowski: **`scheme: http` is hardcoded — breaks HTTPS deployments.**

When `scheme` is set in a `RequestRedirect` filter, Gateway API forces the redirect to that scheme regardless of the incoming request. A user on `https://example.com/` receives a `302 → http://example.com/openmrs/spa/home`, downgrading to plain HTTP. With HSTS this causes an infinite redirect loop.

The README's security table explicitly calls out TLS as the next step for production. Omit the `scheme` field entirely (the redirect will preserve the original scheme) or make it configurable:

```yaml
scheme: {{ .Values.rootRedirect.scheme | default "http" }}
```

helm/openmrs-backend/templates/admin-httproute.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if .Values.seaweedfs.admin.httpRoute.enabled }}
```
rkorytkowski: **Missing `seaweedfs.enabled` and `seaweedfs.admin.enabled` guards.**

This template renders an HTTPRoute pointing to `{{ .Release.Name }}-seaweedfs-admin` when only `seaweedfs.admin.httpRoute.enabled: true` is set. But SeaweedFS is deployed only when `seaweedfs.enabled: true` (the new default in this PR is `false`, gated in `Chart.yaml` via `condition: seaweedfs.enabled`). If the admin HTTPRoute is enabled without the subchart being deployed, the backend service doesn't exist and the gateway returns 503 for all admin paths.

Compare with `grafana-httproute.yaml` in this same PR which correctly AND-gates both flags:
```yaml
{{- if and .Values.monitoring.enabled .Values.grafana.httpRoute.enabled }}
```

This template should use:
```yaml
{{- if and .Values.seaweedfs.enabled .Values.seaweedfs.admin.enabled .Values.seaweedfs.admin.httpRoute.enabled }}
```

helm/openmrs-backend/values.yaml
```diff
@@ -243,8 +256,8 @@ grafana:
   adminPassword: "Admin123"
   grafana.ini: # Needed for serving from path
     server:
-      domain: grafana.local
-      root_url: "%(protocol)s://%(domain)s/grafana"
+      domain: localhost
+      root_url: "http://localhost:8080/grafana"
```
rkorytkowski: **`root_url` hardcoded to `http://localhost:8080/grafana` in the base `values.yaml` — breaks all non-localhost deployments.**

Grafana uses `root_url` to construct OAuth redirect URIs, share links, alert notification URLs, and email links. The previous value `%(protocol)s://%(domain)s/grafana` was environment-agnostic; Grafana interpolates protocol and host from the actual request. The new hardcoded value is only correct for kind-local setups.

Any deployment that does not explicitly override `grafana.grafana\.ini.server.root_url` (staging, CI, production) will have all external Grafana links point to `http://localhost:8080/grafana` — OAuth logins break immediately, share links go nowhere.

Keep the base `values.yaml` using the dynamic form and let `kind-openmrs.yaml` set the hardcoded value:
```yaml
# values.yaml — keep this
root_url: "%(protocol)s://%(domain)s/grafana"

# kind-openmrs.yaml — set the concrete value here
root_url: "http://localhost:8080/grafana"
```

helm/openmrs-backend/values.yaml
```diff
@@ -254,10 +267,20 @@ grafana:
       timeout: 300
       keep_alive_seconds: 300
   ingress:
-    enabled: true
+    enabled: false
```
rkorytkowski, line 270: **`grafana.ingress.enabled` default flipped from `true` to `false` — silent breaking change.**

The base `values.yaml` previously defaulted `grafana.ingress.enabled: true`. Any existing Helm release that relied on the default (without pinning the value) will lose Grafana Ingress access on the next `helm upgrade` with no error or warning. The `HTTPRoute` replacement is also opt-in (`grafana.httpRoute.enabled: false` is the new default), so there is a window after upgrade where Grafana is unreachable via both mechanisms.

Prefer keeping the default unchanged and explicitly disabling ingress only in `kind-openmrs.yaml`. If the intent is to retire the Ingress path, add a deprecation notice in the chart changelog.

rkorytkowski (commented):
This PR adds Gateway API HTTPRoute templates for Grafana and SeaweedFS Admin dashboards, plus a root-to-SPA redirect. Five findings below — two correctness bugs in the new templates, two in the modified values defaults, and one guard inconsistency.

helm/openmrs-backend/values.yaml
rkorytkowski: This is actually correct. We won't follow Claude's suggestion.

suubi-joshua:
Thanks for the review @rkorytkowski
I have adddressed all these in the latest commit.

README.md
```diff
@@ -66,7 +66,16 @@ Once deployment completes, OpenMRS is available at:
...
+With monitoring enabled (default in `kind-openmrs.yaml`), additional dashboards are available at:
+
+| Service | URL |
+|---------|-----|
+| Grafana (logs dashboard) | http://localhost:8080/grafana/ |
+| SeaweedFS Admin (cluster overview & file browser) | http://localhost:8080/seaweedfs-admin/ |
+
+No port-forwarding needed — Traefik binds the port directly. Default credentials: Grafana `admin` / `Admin123`, SeaweedFS Admin `admin` / `Admin123`.
+
+To disable monitoring, set `openmrs-backend.monitoring.enabled=false` in `kind-openmrs.yaml` or pass `--set openmrs-backend.monitoring.enabled=false` to `helm`.
```
rkorytkowski: Please fix docs: openmrs-backend.monitoring.enabled=false doesn't disable seaweedFS, which is correct since it's not monitoring, but docs are incorrect.

rkorytkowski (changes requested):

README.md
suubi-joshua: This skipped my mind but I have now updated the docs, each dashboard in the access table now has a 'Controlled by' column showing its config key, and the disable line explicitly says 'To disable monitoring (Grafana, Loki, Alloy)' to avoid implying SeaweedFS is part of monitoring.

PR #22, opened 2026-07-20
TRUNK 6671:  Add headless service for JGroups DNS_PING and remove infinispan ports from ClusterIP

suubi-joshua:
## Problem
Infinispan cached entries were not replicating across backend replicas. Each pod formed a singleton cluster (view size=1) because JGroups DNS_PING was resolving the ClusterIP service FQDN, which returns a single VIP, not individual pod IPs.

## Three changes to openmrs-backend Helm templates:
1. New `headless-service.yaml` a clusterIP: None Service with `publishNotReadyAddresses: true` so DNS returns one A record per pod, including during startup (before readiness probes pass).
2. `configmap.yaml` point `jgroups.dns.query` from the ClusterIP FQDN to the headless Service FQDN, so DNS_PING discovers all pods.
3. `service.yaml` remove Infinispan ports (7800/7801) from the ClusterIP Service. These were unnecessary (JGroups binds directly on pod IPs) and an unintended exposure.

No `StatefulSet.spec.serviceName` change, the headless Service selects pods by label, avoiding the immutable-field upgrade issue entirely.

## Verification
Tested on a Kind cluster with replicaCount: 2:
- Both pods 1/1 Ready
- DNS getent hosts openmrs-backend-headless.<ns>.svc.cluster.local returns both pod IPs
- Cross-pod TCP connections established on port 7800 (Hibernate cache)
- JGroups cross-cluster message exchange confirmed
- Zero runtime errors
- helm lint passes for both clustered and non-clustered paths

JIRA TICKET: https://openmrs.atlassian.net/browse/TRUNK-6671

### DIFF
--- a/helm/openmrs-backend/templates/configmap.yaml
+++ b/helm/openmrs-backend/templates/configmap.yaml
@@ -48,7 +48,7 @@ data:
   {{- if .Values.infinispan.clustered }}
   cache.type: "cluster"
   cache.stack: "kubernetes"
-  jgroups.dns.query: '{{ include "openmrs-backend.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local'
+  jgroups.dns.query: '{{ include "openmrs-backend.fullname" . }}-headless.{{ .Release.Namespace }}.svc.cluster.local'
   jgroups.bind.port: {{ .Values.infinispan.bind_port | quote }}
   {{- else }}
   cache.type: "local"
--- a/helm/openmrs-backend/templates/configmap.yaml
+++ b/helm/openmrs-backend/templates/configmap.yaml
@@ -48,7 +48,7 @@ data:
   {{- if .Values.infinispan.clustered }}
   cache.type: "cluster"
   cache.stack: "kubernetes"
-  jgroups.dns.query: '{{ include "openmrs-backend.fullname" . }}-headless.{{ .Release.Namespace }}.svc.cluster.local'
+  jgroups.dns.query: '{{ include "openmrs-backend.headlessServiceName" . }}.{{ .Release.Namespace }}.svc.cluster.local'
   jgroups.bind.port: {{ .Values.infinispan.bind_port | quote }}
   {{- else }}
   cache.type: "local"
--- a/helm/openmrs-backend/templates/headless-service.yaml
+++ b/helm/openmrs-backend/templates/headless-service.yaml
@@ -1,0 +1,27 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "openmrs-backend.fullname" . }}-headless
+  labels:
+    {{- include "openmrs-backend.labels" . | nindent 4 }}
+spec:
+  type: ClusterIP
+  clusterIP: None
+  publishNotReadyAddresses: true
+  selector:
+    {{- include "openmrs-backend.selectorLabels" . | nindent 4 }}
+  ports:
+    - name: http
+      port: {{ .Values.service.port }}
+      targetPort: http
+      protocol: TCP
+    {{- if .Values.infinispan.clustered }}
+    - name: infinispan-hibernate
+      port: {{ .Values.infinispan.bind_port }}
+      targetPort: {{ .Values.infinispan.bind_port }}
+      protocol: TCP
+    - name: infinispan-api
+      port: {{ add .Values.infinispan.bind_port 1 }}
+      targetPort: {{ add .Values.infinispan.bind_port 1 }}
+      protocol: TCP
+    {{- end }}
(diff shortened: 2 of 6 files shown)


helm/openmrs-backend/templates/headless-service.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- if .Values.infinispan.clustered }}
+    - name: infinispan-hibernate
+      port: {{ .Values.infinispan.bind_port }}
+      targetPort: {{ .Values.infinispan.bind_port }}
+      protocol: TCP
+    - name: infinispan-api
+      port: {{ add .Values.infinispan.bind_port 1 }}
+      targetPort: {{ add .Values.infinispan.bind_port 1 }}
+      protocol: TCP
+    {{- end }}
```
dkayiwa: The `http` entry gives the namespace a second Service name answering on 8080, and because `publishNotReadyAddresses: true` is set here (rightly, for discovery) that name resolves to pods that are still booting. This chart's startup probe tolerates roughly 20 minutes of boot time, so an HTTPRoute, a probe or a ServiceMonitor aimed at `-headless` by mistake would hand back 502s from a half-started pod with nothing obvious to point at. I'd leave it off.

It isn't buying anything on the discovery side either. Infinispan's kubernetes stack resolves `jgroups.dns.query` as an A record (`dns_record_type="${jgroups.dns.record:A}"` in `default-configs/default-jgroups-kubernetes.xml`, which is what `cache.stack: kubernetes` selects) and takes the peer port from the channel's own `bind_port`, so the SRV records these port entries generate are never looked up. A headless Service publishes the pod A records with or without a `ports` block, which is also why the two infinispan entries here are documentation rather than plumbing.

The ConfigMap only names `-headless` inside the `infinispan.clustered` branch, so gating this file the same way keeps a Service nothing reads out of the default install. I rendered the version below both ways: clustered gives the same two discovery ports and the same FQDN the ConfigMap points at, non-clustered drops the Service entirely and leaves no dangling reference. Neither change blocks merge.

```suggestion
{{- if .Values.infinispan.clustered }}
apiVersion: v1
kind: Service
metadata:
  name: {{ include "openmrs-backend.fullname" . }}-headless
  labels:
    {{- include "openmrs-backend.labels" . | nindent 4 }}
spec:
  type: ClusterIP
  clusterIP: None
  publishNotReadyAddresses: true
  selector:
    {{- include "openmrs-backend.selectorLabels" . | nindent 4 }}
  ports:
    - name: infinispan-hibernate
      port: {{ .Values.infinispan.bind_port }}
      targetPort: {{ .Values.infinispan.bind_port }}
      protocol: TCP
    - name: infinispan-api
      port: {{ add .Values.infinispan.bind_port 1 }}
      targetPort: {{ add .Values.infinispan.bind_port 1 }}
      protocol: TCP
{{- end }}
```

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -48,7 +48,7 @@ data:
   {{- if .Values.infinispan.clustered }}
   cache.type: "cluster"
   cache.stack: "kubernetes"
-  jgroups.dns.query: '{{ include "openmrs-backend.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local'
+  jgroups.dns.query: '{{ include "openmrs-backend.fullname" . }}-headless.{{ .Release.Namespace }}.svc.cluster.local'
```
dkayiwa: The FQDN here and the Service name over in `headless-service.yaml` are built independently, so nothing keeps the two in sync. Edit the suffix on one side later and DNS_PING resolves a name that doesn't exist, each pod forms its own one-member cache, and the readiness probe still passes, which is the same silent failure this PR is fixing. Helm Build wouldn't catch it either: `helm lint` runs with the default `clustered: false` so it never renders either side, and the `helm template` steps only assert that rendering succeeds, which it would.

Pulling the name into `_helpers.tpl` takes the drift off the table, and it's what the chart already does for `elasticsearch.serviceName` and `openmrs-seaweedfs.s3url`:

```
{{/*
Headless Service name used for JGroups DNS_PING discovery.
*/}}
{{- define "openmrs-backend.headlessServiceName" -}}
{{- printf "%s-headless" (include "openmrs-backend.fullname" .) -}}
{{- end }}
```

This line then becomes

```
  jgroups.dns.query: '{{ include "openmrs-backend.headlessServiceName" . }}.{{ .Release.Namespace }}.svc.cluster.local'
```

and the Service's `name:` becomes `{{ include "openmrs-backend.headlessServiceName" . }}`. I rendered that variant with `clustered=true` and it comes out byte-identical to what you have now, so it's purely defensive. Not a suggestion block because it only works once the `_helpers.tpl` part lands with it. Fine to merge as-is if you'd rather keep the diff tight.

PR #26, opened 2026-07-28
TRUNK-6659: Make db name configurable independently and fix external-DB credential key

suubi-joshua:
### Summary 

Hardens the `openmrs-backend` chart for external / bring-your-own databases. Two fixes plus supporting doc and CI changes:

1. Configurable database name for external DBs. Adds db.database. The new openmrs-backend.dbName helper is mariadb.enabled-aware:

Embedded MariaDB (`mariadb.enabled=true`):  the name always comes from `mariadb.auth.database` (what the operator creates). `db.database` is ignored, so the app's connection target can never diverge from the operator-created database.
External DB (`mariadb.enabled=false`): the name comes from `db.database`, falling back to `"openmrs"`.
All four DB-name usages in the `ConfigMap` (`OMRS_DB_NAME` + the galera/replication/single-node JDBC URLs) now route through this one helper, so they can't drift apart. The external path now emits `OMRS_DB_NAME` (it previously emitted none), but only when `db.url` is not set, when a full `db.url` is supplied, the name lives in the URL and a separate `OMRS_DB_NAME` would be inert/misleading, so it's omitted (consistent with how OMRS_DB_HOSTNAME/OMRS_DB_PORT are already gated).

2. External-DB credential key fix. `secret.yaml` read `.Values.db.user`, a key that doesn't exist in `values.yaml;` the correct key is `db.username`. Every external-DB deployment was rendering `OMRS_DB_USERNAME=""` and failing to authenticate, with an error that looked like a connectivity problem.

JIRA TICKET: https://openmrs.atlassian.net/browse/TRUNK-6659


README.md
```diff
@@ -144,6 +144,7 @@ Prepend with the name of the service: `openmrs-backend`, `openmrs-frontend`, `tr
 | Name                                                             | Description                                                                                                            | Default Value                                             |
 |------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
 | `openmrs-backend.db.hostname`                                    | Hostname for OpenMRS DB                                                                                                | `""` or defaults to galera or mariadb hostname if enabled |
+| `openmrs-backend.db.database`                                    | Override the OpenMRS database name. Falls back to `mariadb.auth.database` then `"openmrs"`. When embedded MariaDB is enabled, must match `mariadb.auth.database` | `""` |
```
dkayiwa: This row doesn't describe what the helper actually does, and it needs correcting before this lands. `openmrs-backend.dbName` never falls back from `db.database` to `mariadb.auth.database`: with embedded MariaDB it reads `mariadb.auth.database` and ignores `db.database` completely, and with `mariadb.enabled=false` an empty `db.database` goes straight to `"openmrs"`. I rendered this branch with `--set mariadb.enabled=true --set db.database=from_db_key` and the ConfigMap and Secret came out byte-identical to main, so the value really is inert in the embedded case rather than "must match".

If it merges like this, an operator on embedded MariaDB reads the row, sets `openmrs-backend.db.database=openmrs_prod`, and the release still comes up on `openmrs`. Nothing fails, nothing warns, and the rendered manifests give no hint the value was dropped. The values.yaml comment added in this same PR ("Ignored when mariadb.enabled=true") is the accurate one, so it's this row that needs to change.

```suggestion
| `openmrs-backend.db.database`                                    | OpenMRS database name for external (bring-your-own) databases. Empty falls back to `"openmrs"`. Ignored when `mariadb.enabled=true`, where the name always comes from `mariadb.auth.database` | `""` |
```

The PR description states the same inverted precedence (`db.database` -> `mariadb.auth.database` -> `"openmrs"`), so that's worth fixing too, otherwise the merge commit preserves it. One optional extra while you're in this table: `db.username` has no row at all, even though it's the key you fix in secret.yaml, and that gap is plausibly how `db.user` got written in the first place.

helm/openmrs-backend/templates/configmap.yaml
```diff
@@ -7,19 +7,20 @@ metadata:
...
+  OMRS_DB_URL: "jdbc:mariadb:replication://{{ .Release.Name }}-mariadb-primary.{{ .Release.Namespace }}.svc.cluster.local:3306,{{ .Release.Name }}-mariadb-secondary.{{ .Release.Namespace }}.svc.cluster.local:3306/{{ include "openmrs-backend.dbName" . }}?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true"
   {{- else }}
   OMRS_DB_HOSTNAME: "{{ .Release.Name }}-mariadb"
-  OMRS_DB_URL: "jdbc:mariadb://{{ .Release.Name }}-mariadb.{{ .Release.Namespace }}.svc.cluster.local:3306/{{ .Values.mariadb.auth.database | default "openmrs" }}?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true"
+  OMRS_DB_URL: "jdbc:mariadb://{{ .Release.Name }}-mariadb.{{ .Release.Namespace }}.svc.cluster.local:3306/{{ include "openmrs-backend.dbName" . }}?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true"
   {{- end }}
   {{- else }}
+  OMRS_DB_NAME: {{ include "openmrs-backend.dbName" . | quote }}
```
dkayiwa: `OMRS_DB_NAME` is written even when `db.url` is set, and in that case the container ignores it. The image's `startup-init.sh` only assembles a URL from the parts when `OMRS_DB_URL` is empty:

```
OMRS_DB_URL="${OMRS_DB_URL:-jdbc:${OMRS_DB_JDBC_PROTOCOL}://${OMRS_DB_HOSTNAME}:${OMRS_DB_PORT}/${OMRS_DB_NAME}${OMRS_DB_ARGS}${OMRS_DB_EXTRA_ARGS}}"
```

Rendering with `--set db.database=openmrs_rds --set db.url=jdbc:mysql://rds.example.com:3306/other_db` gives a ConfigMap advertising `OMRS_DB_NAME: "openmrs_rds"` while the app connects to `other_db`, which is an unhelpful thing to hand whoever is debugging that connection later. It's also why `OMRS_DB_HOSTNAME` and `OMRS_DB_PORT` a few lines down sit inside the `else`, and the same treatment keeps this key consistent with them:

```suggestion
  {{- if not .Values.db.url }}
  OMRS_DB_NAME: {{ include "openmrs-backend.dbName" . | quote }}
  {{- end }}
```

I rendered that patch to check it: key present with no `db.url`, gone when `db.url` is set, embedded path unchanged, `helm lint` clean. Your call though, since the app connects to the right database either way.

helm/openmrs-backend/templates/secret.yaml
```diff
@@ -10,7 +10,7 @@ stringData:
   OMRS_DB_USERNAME: {{ .Values.mariadb.auth.username | default "openmrs" | quote }}
   OMRS_DB_PASSWORD: {{ .Values.mariadb.auth.password | default "OpenMRS123" | quote }}
   {{ else }}
-  OMRS_DB_USERNAME: {{ .Values.db.user | quote }}
+  OMRS_DB_USERNAME: {{ .Values.db.username | quote }}
```
dkayiwa, line 13: Nothing in CI renders this branch, which is how the `db.user` typo lasted this long. All three `helm template` cases in `helm-build.yaml` run with `mariadb.enabled=true` (the umbrella values set it, `kind-openmrs.yaml` sets it, and the minimal case leaves it alone), so the `mariadb.enabled=false` half of this file and of configmap.yaml is never rendered. Adding a plain render case wouldn't have caught it either, since `helm template` renders a values key that doesn't exist as empty instead of failing, so the check has to assert on the output:

```yaml
- name: Validate Rendering (external database)
  run: |
    helm template helm/openmrs --generate-name \
      --set openmrs-backend.mariadb.enabled=false \
      --set openmrs-backend.db.database=openmrs_rds \
      --set openmrs-backend.db.username=rdsuser > /tmp/external-db.yaml
    grep -q 'OMRS_DB_USERNAME: "rdsuser"' /tmp/external-db.yaml
    grep -q 'OMRS_DB_NAME: "openmrs_rds"' /tmp/external-db.yaml
```

I ran that against both branches: on main both greps fail (`OMRS_DB_USERNAME` renders empty, `OMRS_DB_NAME` is absent), on this branch both pass, so it pins exactly the two things you changed here. Happy for it to be a follow-up if you'd rather keep this PR to the two fixes.

README.md
suubi-joshua: Thanks i forgot to commit this

helm/openmrs-backend/templates/secret.yaml
suubi-joshua: Your are right. 
I'd rather land it here than defer it,  the whole point is that this PR is what introduces the external-DB path, so the coverage should ship with it.

helm/openmrs-backend/templates/configmap.yaml
suubi-joshua: I think this tightens things up, a rendered `OMRS_DB_NAME` the app ignores is a debugging trap, even if the connection itself is fine.

PR #28, opened 2026-08-03
TRUNK-6736: Make openmrs-backend/openmrs-frontend infra-free shared charts

suubi-joshua:
## Summary

`openmrs-backend` currently bundles its own infra (MariaDB CR, SeaweedFS, Grafana/Loki/Alloy, ECK Elasticsearch) as embedded dependencies, which blocks a tenant chart from reusing it without also inheriting a full copy of the primary's infrastructure.

This PR makes `openmrs-backend`/`openmrs-frontend` **workload-only** charts and moves the infra up into the `openmrs` umbrella, which now deploys it directly. A tenant chart can depend on the two shared charts with no infra flags to disable. 

## Key changes
`openmrs-backend` is now infra-free. Removed the MariaDB, SeaweedFS, Grafana/Loki/Alloy, and Elasticsearch templates and dependencies. The chart now renders only the application's own resources (StatefulSet, Service, ConfigMap, Secret, routes).
The `openmrs` umbrella now owns that infrastructure directly, deploying the same MariaDB/monitoring/storage/search stack as before, just from the umbrella instead of the backend.

Shared credentials have one source of truth. The database username/password used to be declared twice (once for the actual MariaDB, once for the app connecting to it) with just a comment reminding you to keep them in sync. They're now a single value both sides read automatically.
The frontend's SPA settings are now configurable (URL paths, default locale, etc.) instead of hardcoded, so a future tenant can customize them without forking the chart.

CI now actually checks the database connection renders correctly for the normal (embedded database) setup and previously only the "external database" test case was checked, which is exactly the path this change touches most.
Chart versions are no longer forced to move in lockstep on release. `openmrs-backend` and `openmrs-frontend` can now be released on their own schedule, independent of the umbrella and important once a tenant chart depends on a specific version of them.

Related PR: https://github.com/openmrs/openmrs-contrib-cluster/pull/19
JIRA Ticket: https://openmrs.atlassian.net/browse/TRUNK-6736


helm/openmrs/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.3.0
+version: 1.4.0
```
dkayiwa: This can't go out as a minor bump, because the value moves are silent. I rendered this chart against `main`'s `kind-openmrs.yaml` with no edits: exit 0, no warning, and the output drops from 78 resources to 20. The alloy, grafana, loki, seaweedfs and elasticsearch-eck subcharts all vanish, because `openmrs-backend.monitoring.*`, `.grafana.*`, `.loki.*`, `.alloy.*`, `.elasticsearch-eck.*` and most of `.seaweedfs.*` are now keys nothing reads, and Helm ignores unknown values. What survives makes it worse: the backend ConfigMap still carries `storage.type: s3` pointing at `openmrs-seaweedfs-s3`, `OMRS_SEARCH_ES_URIS` pointing at `openmrs-elasticsearch-eck-es-http`, and an `openmrs-seaweedfs-admin` HTTPRoute is still emitted for a Service that is gone. `openmrs-backend.mariadb.auth.password` is in the same bucket and silently reverts to the `OpenMRS123` default. So anyone who upgrades a 1.3.0 release using their own values file loses monitoring and object storage and ends up with a backend wired to endpoints that no longer resolve, with nothing in the output saying so.

Both charts are published for outside use (`openmrs-1.3.0` and `openmrs-backend-1.3.0` are in the repo index), so I'd make this 2.0.0 and fail the render on the removed keys instead of dropping them. I tried this guard: it renders clean on the new values file and stops on the old one.

```
{{- range list "monitoring" "grafana" "loki" "alloy" "elasticsearch-eck" "galera" }}
{{- if hasKey (index $.Values "openmrs-backend") . }}
{{- fail (printf "openmrs-backend.%s moved to the umbrella top level; see the README migration notes" .) }}
{{- end }}
{{- end }}
```

A short migration section in the README mapping old key to new would cover the rest.

helm/openmrs/values.yaml
```diff
@@ -14,29 +21,253 @@ openmrs-backend:
...
+      size: 8Gi
+    resources:
+      requests:
+        cpu: 500m
+        memory: 512Mi
+      limits:
+        cpu: "1"
+        memory: 1Gi
+  replicas: 2
+  galera: true
```
dkayiwa: `galera` and `replicas` now live in two places with nothing keeping them in sync, and that needs sorting before this merges. The MariaDB CR reads these; the JDBC URL in `openmrs-backend/templates/configmap.yaml` reads `openmrs-backend.mariadb.galera`/`.replicas`, which still default to `true`/`2` in the subchart. The command the README gives at line 110 (`--set mariadb.galera=false --set mariadb.replicas=2`) therefore renders a CR with `replication.enabled: true` alongside a backend still on `jdbc:mariadb:loadbalance://release-name-mariadb...`. I rendered both branches to check: on `main`, the single `openmrs-backend.mariadb.galera=false` flipped the pair together and gave the `replication://...-primary,...-secondary` URL, so this is new.

It breaks at runtime rather than at render time. In mariadb-operator 26.6.0 the plain `<name>` Service is built by `reconcileDefaultService` with role-agnostic selector labels, so it fans out across the primary and the replicas, and the replication controller puts replicas into `read_only=1`. Connections landing on a replica fail their writes with a read-only error while the rest succeed, which is exactly why the non-Galera branch of that ConfigMap targets `-primary`/`-secondary`.

`mariadb.enabled` splits the same way: `--set mariadb.enabled=false` on its own drops the CR and its Secret and leaves the backend pointing at `release-name-mariadb`. Moving `enabled`, `galera` and `replicas` under `global.mariadb` next to `auth` fixes both, using the mechanism this PR already introduced. Worth a CI case as well, asserting the JDBC URL matches the CR mode, since the greps added here only exercise the Galera default.

helm/openmrs-frontend/values.yaml
```diff
@@ -2,8 +2,17 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+nameOverride: "frontend"
+
 enabled: true
 
+# SPA runtime configuration (rendered into the frontend ConfigMap). Defaults match
+# the primary umbrella's values; a tenant overrides per-tenant (e.g. configUrls).
+spaPath: "/openmrs/spa"
```
dkayiwa, line 9: The knob is only half wired, though it's fine to leave for a follow-up. `spaPath` reaches the ConfigMap, but the HTTPRoute match comes from `gateway.path`, which hardcodes `/openmrs/spa` separately in `templates/httproute.yaml:17`, as do `rootRedirect.targetPath` and the ingress rule. Rendering with `--set openmrs-frontend.spaPath=/tenant-a/spa` gives `SPA_PATH: "/tenant-a/spa"` next to a route still matching `/openmrs/spa`, so the tenant case this is meant to unblock still needs both set. Defaulting the route to the new value, `{{ .Values.gateway.path | default .Values.spaPath }}` and likewise for `rootRedirect.targetPath`, would make one setting enough.

helm/openmrs-backend/values.yaml
```diff
@@ -2,12 +2,21 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+nameOverride: "backend"
```
dkayiwa: Did you mean this to rename standalone installs too? `openmrs-backend` is published on its own (it's in the repo index at 1.3.0), and with this default every resource goes from `<release>-openmrs-backend` to `<release>-backend`, StatefulSet included, so an in-place upgrade of a standalone install recreates it and strands the old PVC. The umbrella already sets `nameOverride: backend` in its own values, so dropping this line leaves the umbrella output byte-identical and lets the consumer decide. I'd drop it unless the tenant chart in #19 relies on the default.

helm/openmrs/Chart.yaml
suubi-joshua: You're right, thanks @dkayiwa Confirmed the same result on my end. Let me fix it also I am going to do a major release once this PR gets in.

dkayiwa: The guard doesn't cover the two flags that kept their path, and I think that has to be closed before this goes out.

On `main`, `--set openmrs-backend.elasticsearch.enabled=true` renders an Elasticsearch CR and `--set openmrs-backend.seaweedfs.enabled=true` renders 16 SeaweedFS resources, because both were the `condition:` on the subchart dependencies in the old `openmrs-backend/Chart.yaml`. On this branch the same two flags render zero of either, exit 0, and the backend still gets `OMRS_SEARCH_ES_URIS: https://<release>-elasticsearch-eck-es-http...` and `storage.s3.endpoint: http://<release>-seaweedfs-s3...`. `hasKey` can't see them because the paths didn't change, and README.md:144-147 files both under "Unchanged", which holds for 2.0.0 but not for the 1.x someone is migrating from.

What that costs: a 1.x user works through the guard's errors one at a time, fixes `monitoring`, fixes `mariadb`, gets a clean render and ships it. The backend pod then never starts, because the StatefulSet still mounts `<release>-elasticsearch-eck-es-http-certs-public` and `<release>-elasticsearch-eck-es-elastic-user` with no `optional: true`, and runs `wait-for-elasticsearch` against a CR nobody creates. If they only had SeaweedFS on, it comes up instead with `storage.type: s3` pointing at a Service that doesn't resolve and attachments fail at runtime. The clean render is what makes them trust it.

Inside the umbrella neither combination is ever right, since `openmrs-seaweedfs.s3url` and `openmrs-elasticsearch.url` are both hardcoded to `{{ .Release.Name }}-...`, so a cross-check in the same NOTES.txt block is enough. I ran this one: clean on defaults, on `kind-openmrs.yaml` and on the minimal CI case, and it stops on both 1.x patterns.

```
{{- $backend := index .Values "openmrs-backend" }}
{{- if and $backend.seaweedfs.enabled (not .Values.seaweedfs.enabled) }}
{{- fail "openmrs-backend.seaweedfs.enabled only wires the workload in 2.0.0; set seaweedfs.enabled=true to deploy SeaweedFS, see README.md" }}
{{- end }}
{{- if and $backend.elasticsearch.enabled (not .Values.elasticsearch.enabled) }}
{{- fail "openmrs-backend.elasticsearch.enabled only wires the workload in 2.0.0; set elasticsearch.enabled=true to deploy the ECK cluster, see README.md" }}
{{- end }}
```

The "Unchanged" note wants the other half as well: the path stayed put, but you now have to add the matching top-level flag.

.github/workflows/helm-build.yaml
```diff
@@ -46,25 +46,40 @@ jobs:
...
-            --set openmrs-backend.mariadb.enabled=false \
+            --set global.mariadb.enabled=false \
             --set openmrs-backend.db.database=openmrs_rds \
             --set openmrs-backend.db.username=rdsuser > /tmp/external-db.yaml
           grep -q 'OMRS_DB_USERNAME: "rdsuser"' /tmp/external-db.yaml
-          grep -q 'OMRS_DB_NAME: "openmrs_rds"' /tmp/external-db.yaml
\ No newline at end of file
+          grep -q 'OMRS_DB_NAME: "openmrs_rds"' /tmp/external-db.yaml
+
+      - name: Validate Rendering (MariaDB replication mode)
```
dkayiwa, line 79: Nothing here asserts the migration guard actually fires, which makes it the one piece of the 2.0.0 story that can rot without anyone noticing. Not a blocker, but worth adding while it's fresh, because it breaks in both directions: delete or reword the `fail` block and 1.x values go quiet again, or re-add a `mariadb:`/`monitoring:` key to `openmrs-backend/values.yaml` and every umbrella render dies, since `hasKey` is reading the coalesced subchart defaults.

A step next to this one covers both. I ran it against this branch (passes) and against `main` (fails, as it should, since there's no guard there):

```yaml
      - name: Validate 1.x values are rejected
        run: |
          if helm template helm/openmrs --generate-name \
               --set openmrs-backend.monitoring.enabled=true >/dev/null 2>&1; then
            echo "expected the 2.0.0 guard to reject openmrs-backend.monitoring.*"; exit 1
          fi
          if helm template helm/openmrs --generate-name \
               --set openmrs-backend.mariadb.auth.password=x >/dev/null 2>&1; then
            echo "expected the 2.0.0 guard to reject openmrs-backend.mariadb.*"; exit 1
          fi
```

If you add the seaweedfs/elasticsearch cross-check from the other thread, two more `if` blocks in the same shape cover that too.

README.md
```diff
@@ -131,7 +156,7 @@ To use a MariaDB Galera cluster instead of basic primary-secondary replication:
 
 #### Common parameters
 
-Prepend with the name of the service: `openmrs-backend`, `openmrs-frontend`, `traefik-gateway`, `openmrs-backend.mariadb`, `openmrs-backend.galera`.
+Prepend with the name of the service: `openmrs-backend`, `openmrs-frontend`, `traefik-gateway`, `openmrs-backend.mariadb`.
```
dkayiwa: `openmrs-backend.mariadb` is one of the prefixes the new NOTES.txt guard hard-fails on, so someone following this line and running `--set openmrs-backend.mariadb.image.tag=11.4` now stops the render instead of setting the tag. The MariaDB image moved to the umbrella's own top-level `mariadb.image.*`.

```suggestion
Prepend with the name of the service: `openmrs-backend`, `openmrs-frontend`, `traefik-gateway`, `mariadb`.
```

suubi-joshua: Good catch, applied your suggestion exactly.

helm/openmrs/Chart.yaml
suubi-joshua: You're right and confirmed the same failure mode.

Found the same trap on `openmrs-backend.seaweedfs.admin.enabled` (used to deploy the Admin StatefulSet, now only wires the HTTPRoute) and on `openmrs-backend.seaweedfs.master.*/.volume.*/.filer.*/.s3.enabled/.s3.replicas/.s3.enableAuth/.admin.ingress.*` (moved to the top level like monitoring/grafana, but never added to the guard).

Guarded all of them the same way. Verified against the pre-refactor base and each renders clean-but-wrong before the fix, fails loud after. No false positives on defaults, `kind-openmrs.yam`l, minimal, external-db, or replication-mode.

PR #31, opened 2026-08-17
TRUNK-6653: Add automated DB provisioning and render-time guards to openmrs-tenant chart.

suubi-joshua:
## Summary
Automates tenant database/user/grant provisioning via MariaDB operator CRs (`dbBootstrap.enabled=true` by default), eliminating manual SQL steps. Adds render-time guards that reject unsafe credentials (CWE-89: '/\ in username, password, or database name) and catch `openmrs-backend.db.url` ↔ `dbBootstrap.database` mismatches before install. CI negative tests cover all four guards.

Key files: `db-bootstrap-{database,user,grant,secret}.yaml` (new CRs),` _helpers.tpl` (labels + dbName helper), `validate-db.yaml` (guards), `values.yaml` (dbBootstrap section), `README` and CI updates.

JIRA Ticket: https://openmrs.atlassian.net/browse/TRUNK-6653

### DIFF
--- a/.github/workflows/helm-build.yaml
+++ b/.github/workflows/helm-build.yaml
@@ -135,7 +135,7 @@ jobs:
           TENANT_VALUES="--set global.tenant.name=tenant1 \
             --set openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/openmrs_tenant1?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true \
             --set openmrs-backend.db.hostname=mariadb.example.svc.cluster.local \
-            --set openmrs-backend.db.username=ci_user \
+            --set openmrs-backend.db.username=openmrs_tenant1_user \
             --set openmrs-backend.db.password=ci-pass"
           helm lint helm/openmrs-tenant $TENANT_VALUES
           helm template tenant1 helm/openmrs-tenant $TENANT_VALUES > /tmp/tenant.yaml
@@ -144,19 +144,64 @@ jobs:
           grep -q 'OMRS_DB_URL: "jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/openmrs_tenant1?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true"' /tmp/tenant.yaml
           grep -q 'OMRS_DB_HOSTNAME: "mariadb.example.svc.cluster.local"' /tmp/tenant.yaml
           grep -q 'SPA_PATH: "/openmrs/spa"' /tmp/tenant.yaml
-          [ "$(grep -c 'app.kubernetes.io/tenant: "tenant1"' /tmp/tenant.yaml)" -eq 2 ]
+          # Backend and frontend pod templates (at least) carry the tenant label.
+          [ "$(grep -c 'app.kubernetes.io/tenant: "tenant1"' /tmp/tenant.yaml)" -ge 6 ]
           if grep -q 'storageClassName: ""' /tmp/tenant.yaml; then
             echo "empty storageClassName means no-class (PVC stays Pending); it must be omitted"; exit 1
           fi
-          # Phase-1 invariant: tenant routing is deferred to Phase 3 (TRUNK-6654), so the
-          # shared charts' HTTPRoutes are disabled by default — a host-less route would
-          # collide with the primary and other tenants on the shared gateway.
+          # Phase-1 invariant: HTTPRoutes stay off (host-less routes would collide
+          # on the shared gateway) until Phase 3 (TRUNK-6654).
           if grep -q 'kind: HTTPRoute' /tmp/tenant.yaml; then
             echo "expected NO HTTPRoutes in the default tenant render (gateway must stay off until Phase 3)"; exit 1
           fi
-          # DB connection is required: a tenant without db.url/hostname/username/password
-          # must fail to render, not silently fall back to the shared backend's
-          # primary-DB defaults (host "mariadb"/db "openmrs").
+          # Phase 2: operator Database/User/Grant CRs + credential Secret render,
+          # DB defaults to openmrs_<tenant>.
+          grep -q 'kind: Database' /tmp/tenant.yaml
+          grep -q 'kind: User' /tmp/tenant.yaml
+          grep -q 'kind: Grant' /tmp/tenant.yaml
+          grep -q 'name: tenant1-db-credentials' /tmp/tenant.yaml
+          grep -q 'k8s.mariadb.com/watch: ""' /tmp/tenant.yaml
+          grep -q 'mariaDbRef:' /tmp/tenant.yaml
+          grep -q 'name: "openmrs-mariadb"' /tmp/tenant.yaml
+          grep -q 'namespace: "openmrs"' /tmp/tenant.yaml
+          grep -q 'name: "openmrs_tenant1"' /tmp/tenant.yaml
+          grep -q 'maxUserConnections: 20' /tmp/tenant.yaml
+          [ "$(grep -c 'cleanupPolicy: "Skip"' /tmp/tenant.yaml)" -eq 3 ]
+          grep -q 'characterSet: utf8mb4' /tmp/tenant.yaml
+          grep -q 'collate: utf8mb4_general_ci' /tmp/tenant.yaml
+          # Hyphenated tenants normalize to '_' in the DB name (JDBC URL convention).
+          helm template east-coast helm/openmrs-tenant \
+            --set global.tenant.name=east-coast \
+            --set openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example:3306/openmrs_east_coast \
+            --set openmrs-backend.db.hostname=mariadb.example \
+            --set openmrs-backend.db.username=openmrs_east_coast_user \
+            --set openmrs-backend.db.password=ec-pass > /tmp/tenant-hyphen.yaml
+          grep -q 'name: "openmrs_east_coast"' /tmp/tenant-hyphen.yaml
+          # Grant database must escape _ (SQL LIKE wildcard) to prevent cross-tenant access.
+          # The YAML file contains double backslashes (\\_) because | quote escapes \.
+          grep -q 'database: "openmrs\\\\_east\\\\_coast"' /tmp/tenant-hyphen.yaml
+          # All-digit tenant name: toString prevents int64 render errors.
+          helm template num1 helm/openmrs-tenant $TENANT_VALUES \
+            --set global.tenant.name=12345 \
+            --set openmrs-backend.db.username=openmrs_12345_user \
+            --set openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example:3306/openmrs_12345 > /tmp/tenant-digit.yaml
+          grep -q 'name: "openmrs_12345"' /tmp/tenant-digit.yaml
+          grep -q 'app.kubernetes.io/tenant: "12345"' /tmp/tenant-digit.yaml
+          # dbBootstrap.database override: custom name must appear in CRs and match the URL.
+          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+            --set dbBootstrap.database=openmrs_tenant1_custom \
+            --set "openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/openmrs_tenant1_custom?autoReconnect=true" > /tmp/tenant-override.yaml
+          grep -q 'name: "openmrs_tenant1_custom"' /tmp/tenant-override.yaml
+          # Drift guard: override must still match the URL's database.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set dbBootstrap.database=openmrs_tenant1_custom 2>&1); then
+            echo "expected the guard to reject a dbBootstrap.database override that differs from the URL's database"; exit 1
+          fi
+          grep -qF 'openmrs-tenant: openmrs-backend.db.url must target the same database' <<<"$err"
+          # A '/' inside the JDBC properties must not be read as the database segment.
+          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+            --set "openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/openmrs_tenant1?useSsl=true&serverSslCert=/etc/ssl/certs/mariadb-ca.pem" > /dev/null
+          # Without these the backend silently falls back to the primary DB defaults.
           if err=$(helm template tenant1 helm/openmrs-tenant --set global.tenant.name=tenant1 2>&1); then
             echo "expected the db guard to reject a tenant with no database configured"; exit 1
           fi
@@ -165,6 +210,102 @@ jobs:
             echo "expected the guard to reject a tenant with no global.tenant.name"; exit 1
           fi
           grep -qF 'openmrs-tenant: global.tenant.name is required' <<<"$err"
+          # Tenant name must not contain _ — both - and _ normalize to _, so
+          # east-coast and east_coast would collide on the same MariaDB account.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set global.tenant.name=east_coast 2>&1); then
+            echo "expected the guard to reject a tenant name containing underscores"; exit 1
+          fi
+          grep -qF 'global.tenant.name must not contain underscores' <<<"$err"
+          # Tenant-scoping guard: username must be exactly openmrs_<tenant>_user.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set openmrs-backend.db.username=openmrs 2>&1); then
+            echo "expected the guard to reject a username that is not exactly openmrs_<tenant>_user"; exit 1
+          fi
+          grep -qF 'openmrs-backend.db.username' <<<"$err"
+          # A tenant whose name is a prefix of another's must not reach that tenant's account.
+          if err=$(helm template east helm/openmrs-tenant \
+              --set global.tenant.name=east \
+              --set openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example:3306/openmrs_east \
+              --set openmrs-backend.db.hostname=mariadb.example \
+              --set openmrs-backend.db.username=openmrs_east_coast_user \
+              --set openmrs-backend.db.password=ec-pass 2>&1); then
+            echo "expected the guard to reject another tenant's user"; exit 1
+          fi
+          grep -qF 'must be exactly openmrs_east_user' <<<"$err"
+          # Tenant-scoping guard: database must contain the tenant name. The URL
+          # agrees with the override here, so only the tenant-scoping guard can
+          # be what rejects it.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set dbBootstrap.database=openmrs \
+              --set "openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/openmrs?autoReconnect=true" 2>&1); then
+            echo "expected the guard to reject a database name that does not contain the tenant name"; exit 1
+          fi
+          grep -qF 'dbBootstrap.database (openmrs) must contain the tenant name' <<<"$err"
+          # dbBootstrap disabled → no CRs/Secret leak.
+          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+            --set dbBootstrap.enabled=false > /tmp/tenant-manual.yaml
+          if grep -q 'db-bootstrap' /tmp/tenant-manual.yaml; then
+            echo "expected NO dbBootstrap resources when disabled"; exit 1
+          fi
+          # No imperative bootstrap: no Job, no rootPassword, no SQL or bash
+          # interpolation reaching the database.
+          if grep -qE 'kind: Job|IDENTIFIED BY|CREATE DATABASE|rootPassword|db-bootstrap-job' /tmp/tenant.yaml; then
+            echo "expected NO imperative bootstrap (Job/SQL/rootPassword) in the operator-CRD render"; exit 1
+          fi
+          # Operator CWE-89 (26.6.0): ' and \ in CR-built SQL fields fail render loudly.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set "openmrs-backend.db.password=O'Brien\$(id)" 2>&1); then
+            echo "expected the dbBootstrap guard to reject a password with a single quote"; exit 1
+          fi
+          grep -qF 'openmrs-backend.db.password must not contain single quotes or backslashes' <<<"$err"
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set 'openmrs-backend.db.password=C:\\windows' 2>&1); then
+            echo "expected the dbBootstrap guard to reject a password with a backslash"; exit 1
+          fi
+          grep -qF 'openmrs-backend.db.password must not contain single quotes or backslashes' <<<"$err"
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set "openmrs-backend.db.username=openmrs_tenant1_user'" 2>&1); then
+            echo "expected the dbBootstrap guard to reject a username with a single quote"; exit 1
+          fi
+          grep -qF 'openmrs-backend.db.username must not contain single quotes or backslashes' <<<"$err"
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set 'openmrs-backend.db.username=openmrs_tenant1\\user' 2>&1); then
+            echo "expected the dbBootstrap guard to reject a username with a backslash"; exit 1
+          fi
+          grep -qF 'openmrs-backend.db.username must not contain single quotes or backslashes' <<<"$err"
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set "dbBootstrap.database=openmrs_tenant1't" 2>&1); then
+            echo "expected the guard to reject a database name with a single quote"; exit 1
+          fi
+          grep -qF 'single quotes, backslashes or backticks' <<<"$err"
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set "dbBootstrap.database=openmrs_tenant1\`db" 2>&1); then
+            echo "expected the dbBootstrap guard to reject a database name with a backtick"; exit 1
+          fi
+          grep -qF 'single quotes, backslashes or backticks' <<<"$err"
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set 'dbBootstrap.database=openmrs_tenant1\\db' 2>&1); then
+            echo "expected the guard to reject a database name with a backslash"; exit 1
+          fi
+          grep -qF 'single quotes, backslashes or backticks' <<<"$err"
+          # The URL's database segment must match the provisioned database — they are the same DB.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set "openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/other_db?autoReconnect=true" 2>&1); then
+            echo "expected the guard to reject a db.url whose database differs from dbBootstrap's"; exit 1
+          fi
+          grep -qF 'openmrs-tenant: openmrs-backend.db.url must target the same database' <<<"$err"
+          # Shell metachars without ' or \ still render.
+          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+            --set 'openmrs-backend.db.password=O$Brien$(id)' > /tmp/tenant-inject.yaml
+          grep -qF 'password: "O$Brien$(id)"' /tmp/tenant-inject.yaml
+          if grep -qE 'IDENTIFIED BY|CREATE DATABASE|DB_PASS=|mariadb -u' /tmp/tenant-inject.yaml; then
+            echo "expected no shell/SQL interpolation surface in the operator-CRD render"; exit 1
+          fi
+          # All-digit password: quote handles int64 without toString.
+          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+            --set 'openmrs-backend.db.password=12345' > /tmp/tenant-digit-pw.yaml
+          grep -qF 'password: "12345"' /tmp/tenant-digit-pw.yaml
           # Forward check (Phase 3): when routing IS enabled, the per-tenant spaPath still
           # reaches the route match and the redirect target, not just the ConfigMap.
           helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
--- a/README.md
+++ b/README.md
@@ -110,16 +110,28 @@ configuration on top of them. Each tenant is a separate Helm release in its own 
 
 - Primary OpenMRS stack deployed and running (steps 1–4 above)
 - MariaDB accessible from tenant namespace (default DNS: `<primary-release>-mariadb.<primary-namespace>.svc.cluster.local`, e.g. `openmrs-mariadb.openmrs.svc.cluster.local`)
-- A database and user created for the tenant:
+- The **MariaDB operator** running **cluster-wide** (installed automatically by step 3's `openmrs-operator` chart; it watches all namespaces and reconciles the `Database` / `User` / `Grant` CRs below)
 
-```bash
-kubectl exec -n openmrs svc/openmrs-mariadb -- mysql -u root -pRoot123 -e "
-  CREATE DATABASE IF NOT EXISTS openmrs_<tenant> CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
-  CREATE USER IF NOT EXISTS '<tenant>_user'@'%' IDENTIFIED BY '<password>';
-  GRANT ALL PRIVILEGES ON openmrs_<tenant>.* TO '<tenant>_user'@'%';
-  FLUSH PRIVILEGES;
-"
-```
+The tenant's database and user are **created automatically**: the tenant chart renders the
+MariaDB operator's `Database` / `User` / `Grant` custom resources (in the tenant namespace,
+referencing the shared MariaDB via `mariaDbRef`). The operator turns them into
+`CREATE DATABASE` / `CREATE USER` / `GRANT` / `ALTER USER ... WITH MAX_USER_CONNECTIONS`
+declaratively — no imperative bootstrap Job runs, and the shared MariaDB's **root password
+never enters the tenant namespaces** (it stays inside the operator). Nothing to do by hand.
+
+> **Manual fallback:** with `--set dbBootstrap.enabled=false` the CRs are not
+> rendered and you must create the database and user yourself, e.g.:
+>
+> ```bash
+> # _ is a LIKE wildcard in GRANT; escape each _ as \_ and wrap in backticks.
+> # A quoted heredoc ('SQL') keeps backticks intact through the local shell.
+> kubectl exec -i -n openmrs svc/openmrs-mariadb -- mysql -u root -pRoot123 <<'SQL'
+>   CREATE DATABASE IF NOT EXISTS openmrs_east_coast CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
+>   CREATE USER IF NOT EXISTS 'openmrs_east_coast_user'@'%' IDENTIFIED BY '<password>';
+>   GRANT ALL PRIVILEGES ON `openmrs\_east\_coast`.* TO 'openmrs_east_coast_user'@'%';
+>   FLUSH PRIVILEGES;
+> SQL
+> ```
 
 #### Install a tenant
 
@@ -140,9 +152,33 @@ helm install <tenant> helm/openmrs-tenant \
   --set openmrs-backend.db.url="jdbc:mariadb:loadbalance://<primary-release>-mariadb.<primary-namespace>.svc.cluster.local:3306/openmrs_<tenant>?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true" \
   --set openmrs-backend.db.hostname=<primary-release>-mariadb.<primary-namespace>.svc.cluster.local \
   --set openmrs-backend.db.port=3306 \
-  --set openmrs-backend.db.username=<tenant>_user \
-  --set openmrs-backend.db.password=<password>
+  --set openmrs-backend.db.username=openmrs_<tenant>_user \
+  --set openmrs-backend.db.password=<password> \
+  --set dbBootstrap.mariaDbRef.name=<primary-release>-mariadb \
+  --set dbBootstrap.mariaDbRef.namespace=<primary-namespace>
 ```
+
+> **Hyphenated tenant names:** the database defaults to `openmrs_<tenant>` with
+> any `-` in the tenant name replaced by `_` (JDBC URL convention). For example,
+> `global.tenant.name=east-coast` provisions `openmrs_east_coast`, so use
+> `.../openmrs_east_coast?...` in the JDBC URL, not `openmrs_east-coast`.
+> **Do not use `_` in tenant names** — both `-` and `_` normalize to `_`, so
+> `east-coast` and `east_coast` would collide on the same MariaDB account.
+
+The chart renders the MariaDB operator's `Database` / `User` / `Grant` CRs, which
+auto-provision the tenant database `openmrs_<tenant>` (override via
+`dbBootstrap.database`), the `openmrs_<tenant>_user` user, and its grants — the password is
+taken from `openmrs-backend.db.password` (matching the backend's own Secret). The CRs
+target the shared MariaDB via `dbBootstrap.mariaDbRef` (defaults to the primary chart's
+`openmrs-mariadb` CR in the `openmrs` namespace). Set `--set dbBootstrap.enabled=false`
+to skip bootstrapping and manage the database manually (see above).
+
+> **Known operator limitation:** the MariaDB operator (26.6.0, installed by
+> `openmrs-operator`) has an upstream SQL-injection advisory (CWE-89): it builds SQL
+> without escaping `'` in user/database names or passwords. The tenant chart rejects
+> `'`, `\` and `` ` `` (database name only) while `dbBootstrap.enabled=true` so the install fails
+> loudly instead of leaving a `User`/`Grant` CR stuck in `Error`. Change the value or
+> set `dbBootstrap.enabled=false` and create the user manually.
 
 Example for a tenant named `coast`:
 
@@ -154,7 +190,7 @@ helm install coast helm/openmrs-tenant \
   --set openmrs-backend.db.url="jdbc:mariadb:loadbalance://openmrs-mariadb.openmrs.svc.cluster.local:3306/openmrs_coast?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true" \
   --set openmrs-backend.db.hostname=openmrs-mariadb.openmrs.svc.cluster.local \
   --set openmrs-backend.db.port=3306 \
-  --set openmrs-backend.db.username=coast_user \
+  --set openmrs-backend.db.username=openmrs_coast_user \
   --set openmrs-backend.db.password=CoastPass123
 ```
 
@@ -175,6 +211,9 @@ kubectl get pods -n tenant-<tenant> -L app.kubernetes.io/tenant
 
 # Wait for ready
 kubectl wait --for=condition=ready pod -n tenant-<tenant> --all --timeout=600s
+
+# Check the MariaDB operator reconciled the tenant's CRs (Database/User/Grant ready)
+kubectl get database,user,grant -n tenant-<tenant>
 
 # Port-forward to backend (API)
 kubectl port-forward -n tenant-<tenant> svc/<tenant>-openmrs-backend 8080:8080
@@ -208,8 +247,14 @@ through. For the full shared-chart surface, see `helm/openmrs-backend/values.yam
 | `openmrs-backend.db.url` | Full JDBC URL to the shared MariaDB | **required** |
 | `openmrs-backend.db.hostname` | Shared MariaDB host, used by the backend's `wait-for-it` preflight (`OMRS_DB_HOSTNAME`) | **required** |
 | `openmrs-backend.db.port` | Shared MariaDB port | `3306` |
-| `openmrs-backend.db.username` | Tenant DB user | **required** |
+| `openmrs-backend.db.username` | Tenant DB user; must be exactly `openmrs_<tenant>_user` when `dbBootstrap.enabled=true` | **required** |
 | `openmrs-backend.db.password` | Tenant DB password | **required** |
+| `dbBootstrap.enabled` | Render the MariaDB operator's `Database` / `User` / `Grant` CRs that auto-provision the tenant's DB/user/grants on the shared MariaDB | `true` |
+| `dbBootstrap.mariaDbRef.name` | Name of the shared MariaDB CR to provision on | `openmrs-mariadb` |
+| `dbBootstrap.mariaDbRef.namespace` | Namespace of the shared MariaDB CR | `openmrs` |
+| `dbBootstrap.database` | Tenant database to provision on the shared MariaDB; default `openmrs_<tenant>` with hyphens normalized to underscores. Must contain `openmrs_<tenant>` and be the database in `openmrs-backend.db.url` — a mismatch or missing tenant name fails at render time | `openmrs_<tenant>` |
+| `dbBootstrap.maxUserConnections` | Per-tenant `MAX_USER_CONNECTIONS` limit on the shared MariaDB | `20` |
+| `dbBootstrap.cleanupPolicy` | Whether the DB/user are dropped on `helm uninstall`: `Skip` (safe default, keeps tenant data) or `Delete` | `Skip` |
 | `openmrs-backend.podLabels` | Extra labels for backend pods | `{}` |
 | `openmrs-frontend.enabled` | Deploy the frontend | `true` |
 | `openmrs-frontend.spaPath` | URL path the SPA is served from (`SPA_PATH`) | `"/openmrs/spa"` |
--- a/helm/openmrs-tenant/templates/_helpers.tpl
+++ b/helm/openmrs-tenant/templates/_helpers.tpl
@@ -1,0 +1,23 @@
+{{- /*
+Standard labels for every K8s resource the tenant chart renders.
+Use in metadata.labels via: {{ include "openmrs-tenant.labels" . | nindent N }}
+*/}}
+{{- define "openmrs-tenant.labels" -}}
+app.kubernetes.io/name: openmrs-tenant
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/tenant: {{ .Values.global.tenant.name | quote }}
+{{- end }}
+
+{{- /*
+The tenant-scoped name openmrs_<tenant> (hyphens become underscores so it stays
+consistent with the openmrs_<tenant> JDBC URL convention). Both the default
+database name and the required DB username are built from it.
+*/}}
+{{- define "openmrs-tenant.normalizedTenant" -}}
+{{- printf "openmrs_%s" (.Values.global.tenant.name | toString | replace "-" "_") -}}
+{{- end }}
+
+{{- define "openmrs-tenant.dbName" -}}
+{{- .Values.dbBootstrap.database | default (include "openmrs-tenant.normalizedTenant" .) -}}
+{{- end }}
--- a/helm/openmrs-tenant/templates/db-bootstrap-grant.yaml
+++ b/helm/openmrs-tenant/templates/db-bootstrap-grant.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.dbBootstrap.enabled }}
+apiVersion: k8s.mariadb.com/v1alpha1
+kind: Grant
+metadata:
+  name: {{ .Release.Name }}-grant
+  labels:
+    {{- include "openmrs-tenant.labels" . | nindent 4 }}
+spec:
+  mariaDbRef:
+    name: {{ .Values.dbBootstrap.mariaDbRef.name | quote }}
+    namespace: {{ .Values.dbBootstrap.mariaDbRef.namespace | quote }}
+  privileges:
+    - "ALL PRIVILEGES"
+  database: {{ include "openmrs-tenant.dbName" . | replace "_" "\\_" | quote }}
+  table: "*"
+  username: {{ (index .Values "openmrs-backend").db.username | quote }}
+  host: "%"
+  cleanupPolicy: {{ .Values.dbBootstrap.cleanupPolicy | quote }}
+{{- end }}
--- a/helm/openmrs-tenant/templates/db-bootstrap-secret.yaml
+++ b/helm/openmrs-tenant/templates/db-bootstrap-secret.yaml
@@ -1,0 +1,12 @@
+{{- if .Values.dbBootstrap.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-db-credentials
+  labels:
+    {{- include "openmrs-tenant.labels" . | nindent 4 }}
+    k8s.mariadb.com/watch: ""
+type: Opaque
+stringData:
+  password: {{ (index .Values "openmrs-backend").db.password | quote }}
+{{- end }}
--- a/helm/openmrs-tenant/templates/validate-db.yaml
+++ b/helm/openmrs-tenant/templates/validate-db.yaml
@@ -1,7 +1,6 @@
 {{- /*
-Fail fast if the tenant's shared-MariaDB connection isn't fully set. Left empty,
-the shared backend falls back to its own defaults (host "mariadb", db "openmrs"
-— the primary's) with blank creds, renders clean, then fails at runtime.
+Fail fast: without these the shared backend silently falls back to the primary's
+defaults (host "mariadb", db "openmrs") and only fails at runtime.
 */}}
 {{- if not .Values.global.tenant.name -}}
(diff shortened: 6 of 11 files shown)


suubi-joshua:
@claude review

helm/openmrs-tenant/templates/db-bootstrap-secret.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.dbBootstrap.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-db-credentials
+  labels:
+    {{- include "openmrs-tenant.labels" . | nindent 4 }}
```
dkayiwa, line 7: This Secret needs the `k8s.mariadb.com/watch` label, or a password change never reaches MariaDB. The operator only watches `passwordSecretKeyRef` Secrets that carry it: `api/v1alpha1/user_indexes.go` wraps the Secret watch in `predicate.PredicateWithLabel(metadata.WatchLabel)`, and `WatchLabel` is `k8s.mariadb.com/watch`. Nothing in the User CR derives from the password either, so a password-only upgrade leaves its spec untouched and no reconcile is triggered from that side.

If merged as-is, `helm upgrade <tenant> ... --set openmrs-backend.db.password=<new>` rewrites this Secret and the backend's own Secret. The backend StatefulSet carries `checksum/secret` on its pod template, so its pods roll straight away and come back up using the new password while MariaDB still holds the old one, and the tenant sits in CrashLoopBackOff on an auth failure. It recovers only when the operator's periodic SQL requeue happens to fire, which defaults to 10h plus up to 1h of jitter (`--requeue-sql` / `--requeue-sql-max-offset`, `cmd/controller/main.go:154`), and `helm upgrade` reports success the whole time. With the label the operator reconciles the User right away and issues `ALTER USER ... IDENTIFIED BY`.

The operator's ClusterRole already has `watch` on secrets cluster-wide, so this is the only change needed.

```suggestion
  labels:
    {{- include "openmrs-tenant.labels" . | nindent 4 }}
    k8s.mariadb.com/watch: ""
```

helm/openmrs-tenant/templates/_helpers.tpl
```diff
@@ -0,0 +1,18 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/tenant: {{ .Values.global.tenant.name }}
+{{- end }}
+
+{{- /*
+Default database name: openmrs_<tenant>. Hyphens become underscores so the name
+stays consistent with the openmrs_<tenant> JDBC URL convention.
+*/}}
+{{- define "openmrs-tenant.dbName" -}}
+{{- .Values.dbBootstrap.database | default (printf "openmrs_%s" (.Values.global.tenant.name | replace "-" "_")) -}}
```
dkayiwa: An all-digit tenant name fails the render here rather than working. `--set global.tenant.name=123` gives:

```
Error: template: openmrs-tenant/templates/_helpers.tpl:17:106: executing "openmrs-tenant.dbName" at <"_">: wrong type for value; expected string; got int64
```

Helm's `--set` parses a bare integer as int64 and `replace` refuses it. Same thing at the two `contains` calls in `validate-db.yaml` (`--set openmrs-backend.db.username=12345` fails at line 20, `--set openmrs-backend.db.password=12345` at line 23) and at `b64enc` in `db-bootstrap-secret.yaml`. All three of those render fine on main today, and the message names no value, so there is nothing in it for the user to act on.

```suggestion
{{- .Values.dbBootstrap.database | default (printf "openmrs_%s" (.Values.global.tenant.name | toString | replace "-" "_")) -}}
```

Once `toString` is in those three spots the render succeeds, but the values then want quoting on the way out too, since the new templates interpolate them bare: `app.kubernetes.io/tenant: 123` on line 9 here, plus `User.spec.name` / `Grant.spec.username` and `Database.spec.name` / `Grant.spec.database`. Those are all `type: string` in the CRDs, and label values are always strings, so a YAML integer gets rejected by the API server. The subcharts already use `| quote` for the tenant label, which is incidentally why the workflow's `grep -c 'app.kubernetes.io/tenant: "tenant1"'` still counts exactly 2 with these four new resources added, so the `-eq 2` to `-ge 2` change at helm-build.yaml:148 is not actually picking them up.

helm/openmrs-tenant/templates/validate-db.yaml
```diff
@@ -10,3 +9,26 @@ the shared backend falls back to its own defaults (host "mariadb", db "openmrs"
...
+{{- fail "openmrs-tenant: openmrs-backend.db.username must not contain single quotes or backslashes while dbBootstrap.enabled=true — the MariaDB operator (upstream CWE-89, unfixed in 26.6.0) cannot safely provision such a user. Change the username or set dbBootstrap.enabled=false and create the user manually." -}}
+{{- end -}}
+{{- if or (contains "'" $db.password) (contains "\\" $db.password) -}}
+{{- fail "openmrs-tenant: openmrs-backend.db.password must not contain single quotes or backslashes while dbBootstrap.enabled=true — the MariaDB operator (upstream CWE-89, unfixed in 26.6.0) cannot safely provision such a password. Change the password or set dbBootstrap.enabled=false and create the user manually." -}}
+{{- end -}}
+{{- if or (contains "'" $dbName) (contains "\\" $dbName) -}}
+{{- fail "openmrs-tenant: the tenant database name (dbBootstrap.database, default openmrs_<tenant>) must not contain single quotes or backslashes while dbBootstrap.enabled=true — the MariaDB operator (upstream CWE-89, unfixed in 26.6.0) cannot safely provision such a database. Change the tenant name or dbBootstrap.database, or set dbBootstrap.enabled=false." -}}
+{{- end -}}
```
dkayiwa: Worth adding a backtick to this one. The database name is the only one of the three values that lands in a backtick-quoted SQL context, and it slips through: `pkg/sql/sql.go` builds ``CREATE DATABASE `%s` `` with a plain `Sprintf`, and `escapeWildcard` wraps the Grant's database the same way. So ``--set dbBootstrap.database='ope`nmrs'`` renders straight into `Database.spec.name` and `Grant.spec.database`, the operator then sends ``CREATE DATABASE `ope`nmrs` ``, that is a 1064 syntax error, and the Database and Grant CRs end up stuck in `Error`, which is the exact outcome this guard exists to prevent. Not an injection escalation (the driver runs with multi-statement off, so a `;` payload cannot chain), just the failure mode the guard promises to catch.

Adding `` (contains "`" $dbName) `` to the condition covers it, and the message on line 27 wants "single quotes, backslashes or backticks" to match. I checked that a backtick inside the double-quoted template literal parses fine under the pinned Helm 3.17.4.

README.md
```diff
@@ -144,6 +154,21 @@ helm install <tenant> helm/openmrs-tenant \
   --set openmrs-backend.db.password=<password>
 ```
 
+The chart renders the MariaDB operator's `Database` / `User` / `Grant` CRs, which
+auto-provision the tenant database `openmrs_<tenant>` (override via
```
dkayiwa, line 169: For a hyphenated tenant, taking `openmrs_<tenant>` literally now breaks the install. `--set global.tenant.name=east-coast` provisions `openmrs_east_coast` (the dbName helper turns `-` into `_`), so the URL from the snippet just above at line 150, `.../openmrs_east-coast?...`, trips the new drift guard:

```
Error: ... openmrs-backend.db.url must target the same database dbBootstrap provisions (openmrs_east_coast); the URL's database segment is openmrs_east-coast.
```

The normalization is spelled out in the values table at the bottom, but not in either place where someone actually writes the name: here, and the JDBC URL in the install command. Hyphenated tenant names are the normal case, the workflow's own fixture uses `east-coast`. A parenthetical on this line saying the database is `openmrs_<tenant>` with any `-` replaced by `_`, plus the same note on the URL line, would save people the round trip.

helm/openmrs-tenant/templates/db-bootstrap-secret.yaml
suubi-joshua: My bad nice one @dkayiwa thanks.

README.md
suubi-joshua: I added a note after the install command explaining that - in the tenant name becomes _ in the database name, with the east-coast → openmrs_east_coast example.

helm/openmrs-tenant/templates/_helpers.tpl
dkayiwa: `b64enc` in `db-bootstrap-secret.yaml` is the one spot left without `toString`, and it still turns a working value into a render failure:

```
$ helm template tenant1 helm/openmrs-tenant $TENANT_VALUES --set openmrs-backend.db.password=12345
Error: template: openmrs-tenant/templates/db-bootstrap-secret.yaml:12:63: executing "openmrs-tenant/templates/db-bootstrap-secret.yaml" at <b64enc>: wrong type for value; expected string; got int64
```

If merged as-is, an all-digit password (or `true`/`false`) fails the install with that message, which names no value for the user to act on, and the same `--set` renders fine on main. Only those forms are affected: `--set` keeps `1.5`, and anything with a non-digit in it, as a string, so it is the bare-integer and boolean cases that reach `b64enc` untyped. Line 12 needs the same `toString`:

```
  password: {{ (index .Values "openmrs-backend").db.password | toString | b64enc }}
```

I ran that, and `--set openmrs-backend.db.password=12345` then renders `password: MTIzNDU=`. The all-digit block in the workflow is worth a case too, since it pins only the tenant name today and that is how this one got through.

suubi-joshua: Ohh yaa, good catch, missed `toString` on the `b64enc` path. Fixed and added an all-digit password case to the CI to pin it.

suubi-joshua:
@claude review

helm/openmrs-tenant/templates/validate-db.yaml
```diff
@@ -10,3 +9,26 @@ the shared backend falls back to its own defaults (host "mariadb", db "openmrs"
...
+{{- if or (contains "'" ($db.password | toString)) (contains "\\" ($db.password | toString)) -}}
+{{- fail "openmrs-tenant: openmrs-backend.db.password must not contain single quotes or backslashes while dbBootstrap.enabled=true — the MariaDB operator (upstream CWE-89, unfixed in 26.6.0) cannot safely provision such a password. Change the password or set dbBootstrap.enabled=false and create the user manually." -}}
+{{- end -}}
+{{- if or (contains "'" $dbName) (contains "\\" $dbName) (contains "`" $dbName) -}}
+{{- fail "openmrs-tenant: the tenant database name (dbBootstrap.database, default openmrs_<tenant>) must not contain single quotes, backslashes or backticks while dbBootstrap.enabled=true — the MariaDB operator (upstream CWE-89, unfixed in 26.6.0) cannot safely provision such a database. Change the tenant name or dbBootstrap.database, or set dbBootstrap.enabled=false." -}}
+{{- end -}}
+{{- /* The URL's database segment must be the very database provisioned below — they are the same database. */ -}}
+{{- $urlDb := first (splitList "?" (last (splitList "/" $db.url))) -}}
```
dkayiwa: This takes the last `/`-separated segment before it strips the query string, so any `/` inside the JDBC properties wins over the real database segment. With `?useSsl=true&serverSslCert=/etc/ssl/certs/mariadb-ca.pem` the guard decides the database is `mariadb-ca.pem`:

```
Error: execution error at (openmrs-tenant/templates/validate-db.yaml:32:4): openmrs-tenant: openmrs-backend.db.url must target the same database dbBootstrap provisions (openmrs_tenant1); the URL's database segment is mariadb-ca.pem.
```

The URL is fine, so there is no way to satisfy the guard other than turning `dbBootstrap.enabled` off, and the message points at the wrong thing. `serverSslCert`, `localSocket`, `keyStore` and `trustStore` all take paths (I confirmed the names against mariadb-java-client's own `Configuration` constant pool). Nothing in this repo sets one today, the MariaDB CR has no TLS, so it is not blocking. Swapping the two splits is still strictly better: strip the query first, then take the last path segment.

```suggestion
{{- $urlDb := last (splitList "/" (first (splitList "?" $db.url))) -}}
```

I ran both orderings. The reordered one fixes the `serverSslCert` case, still returns `openmrs_tenant1` for the multi-host `jdbc:mariadb:replication://a-primary...:3306,a-secondary...:3306/openmrs_tenant1?...` form and for the plain `?autoReconnect=true` form, and still rejects `other_db` with the same message.

helm/openmrs-tenant/templates/db-bootstrap-secret.yaml
```diff
@@ -0,0 +1,13 @@
...
+    k8s.mariadb.com/watch: ""
+type: Opaque
+data:
+  # b64enc keeps the raw password out of rendered manifests.
+  password: {{ (index .Values "openmrs-backend").db.password | toString | b64enc }}
```
dkayiwa: The comment doesn't hold: the same render already carries the password in the clear. `helm template` emits the backend subchart's own Secret just above this one, with `stringData: OMRS_DB_PASSWORD: "ci-pass"`, so base64 here hides nothing from anyone reading the output. Worth not leaving that claim in the file, since it's the sort of thing someone leans on when deciding a render is safe to paste into a ticket.

`stringData` is what both other Secret templates in the repo use (`openmrs-backend/templates/secret.yaml`, `openmrs/templates/mariadb-secret.yaml`), and it also makes the `toString` unnecessary: `quote` already handles the bare-integer and boolean forms that `b64enc` rejects, so the whole type-juggling problem goes away rather than being patched at one call site. The operator is happy either way, `refresolver.SecretKeyRef` reads `secret.Data[key]` and the API server fills that in from `stringData`.

```suggestion
stringData:
  password: {{ (index .Values "openmrs-backend").db.password | quote }}
```

Not blocking. If you take it, the two base64 assertions in the workflow move with it (`EXPECTED=$(printf %s 'O$Brien$(id)' | base64 -w0)` at line 244 and the all-digit `EXPECTED_PW` block at 252), since the Secret then renders `password: "..."` verbatim. I checked that `--set openmrs-backend.db.password=12345` renders `password: "12345"` with no `toString`. If you'd rather leave the encoding alone, just fixing the comment covers the part that matters.

helm/openmrs-tenant/templates/validate-db.yaml
suubi-joshua: This is really thorough review, thanks @dkayiwa

suubi-joshua:
@claude review

helm/openmrs-tenant/templates/validate-db.yaml
```diff
@@ -10,3 +9,26 @@ the shared backend falls back to its own defaults (host "mariadb", db "openmrs"
...
+{{- if .Values.dbBootstrap.enabled -}}
+{{- /*
+Why the guard: the MariaDB operator (26.6.0) builds SQL from CR fields without
+escaping ' or \, so those chars would leave a User/Grant CR stuck in Error.
+CRD presence can't be checked at render time (helm template runs without a
+cluster), so a missing operator is a documented README prerequisite.
+*/}}
+{{- $dbName := include "openmrs-tenant.dbName" . -}}
+{{- if or (contains "'" ($db.username | toString)) (contains "\\" ($db.username | toString)) -}}
+{{- fail "openmrs-tenant: openmrs-backend.db.username must not contain single quotes or backslashes while dbBootstrap.enabled=true — the MariaDB operator (upstream CWE-89, unfixed in 26.6.0) cannot safely provision such a user. Change the username or set dbBootstrap.enabled=false and create the user manually." -}}
```
dkayiwa, line 28: This needs a guard before merge: `dbBootstrap` adopts whatever MariaDB account and database these values happen to name, including ones another release already owns, and every CR still reports Ready.

I ran it on a kind cluster (operator 26.6.0 via `openmrs-operator`, primary `openmrs-mariadb` in `openmrs`). Installing a tenant with `--set openmrs-backend.db.username=openmrs`, the primary's DB user straight out of `helm/openmrs/values.yaml`, took that account over. `internal/controller/user_controller.go` finds `UserExists('openmrs','%')` and issues `ALTER USER ... IDENTIFIED BY` instead of failing, so the stored hash went from `*84041D86...` to `*87905FD0...`, `mariadb -u openmrs -pOpenMRS123` started answering `ERROR 1045 (28000): Access denied`, and the tenant's own password came back holding `GRANT ALL PRIVILEGES` on the primary's `openmrs` database. If merged as-is, the primary OpenMRS backend loses its database on the next connection it opens and the tenant has read/write on the primary's data, with nothing anywhere saying so: all three tenant CRs reported `Ready/Created`, and the primary's own `openmrs-mariadb-user` CR stayed `Ready=True, Created`.

The database half is worse with `cleanupPolicy: Delete`. `--set dbBootstrap.database=openmrs` satisfies the drift guard as long as the JDBC URL agrees, and deleting that release dropped the primary's `openmrs` schema. Neither of these could happen before this PR, since the tenant chart issued no DDL and borrowed credentials only meant the tenant itself failed to connect.

What makes me want it fixed rather than documented is that `dbBootstrap.enabled=true` is the default, so the destructive path is the one you get without asking for it, and the only thing steering people off it is the `<tenant>_user` example further down the README.

Cheapest fix that covers both, and it sits naturally next to the three guards above: require the names to be tenant-scoped, so the tenant name normalized the way `openmrs-tenant.dbName` already normalizes it (`| toString | replace "-" "_"`) has to appear in `openmrs-backend.db.username` and in `$dbName`. That also rules out two tenants quietly sharing one database, and `dbBootstrap.enabled=false` stays the way out for anyone who really does want to attach to an account that already exists. It does mean renaming the workflow fixtures, since `ci_user` (tenant `tenant1`), `ec_user` (`east-coast`) and the `custom_db` override would all be rejected as written.

README.md
```diff
@@ -110,16 +110,26 @@ configuration on top of them. Each tenant is a separate Helm release in its own
 
 - Primary OpenMRS stack deployed and running (steps 1–4 above)
 - MariaDB accessible from tenant namespace (default DNS: `<primary-release>-mariadb.<primary-namespace>.svc.cluster.local`, e.g. `openmrs-mariadb.openmrs.svc.cluster.local`)
-- A database and user created for the tenant:
+- The **MariaDB operator** running **cluster-wide** (installed automatically by step 1's `openmrs-operator` chart; it watches all namespaces and reconciles the `Database` / `User` / `Grant` CRs below)
```
dkayiwa: The `openmrs-operator` chart goes in at step 3 of the `make deploy` table above, step 1 being the preflight checks and image pre-pull, so this points at the wrong row for the one prerequisite people come back here to check when the CRs are not reconciling.

```suggestion
- The **MariaDB operator** running **cluster-wide** (installed automatically by step 3's `openmrs-operator` chart; it watches all namespaces and reconciles the `Database` / `User` / `Grant` CRs below)
```

helm/openmrs-tenant/templates/validate-db.yaml
suubi-joshua: Thanks again for the detailed reviews @dkayiwa. As you have suggested, I have added tenant-scoping guards: both `openmrs-backend.db.username` and `dbBootstrap.database` must contain the normalized tenant name (`openmrs_<tenant>`). This prevents one tenant from taking over another's account or database. `dbBootstrap.enabled=false` stays as the escape hatch. I have also updated CI to use tenant-scoped names, plus two new negative tests.

README.md
```diff
@@ -144,6 +154,26 @@ helm install <tenant> helm/openmrs-tenant \
...
 ```
 
+> **Hyphenated tenant names:** the database defaults to `openmrs_<tenant>` with
+> any `-` in the tenant name replaced by `_` (JDBC URL convention). For example,
+> `global.tenant.name=east-coast` provisions `openmrs_east_coast`, so use
+> `.../openmrs_east_coast?...` in the JDBC URL, not `openmrs_east-coast`.
+
+The chart renders the MariaDB operator's `Database` / `User` / `Grant` CRs, which
+auto-provision the tenant database `openmrs_<tenant>` (override via
+`dbBootstrap.database`), the `<tenant>_user` user, and its grants — the password is
```
dkayiwa: This needs a docs fix before merge: the install instructions in this file no longer render. The username guard in `validate-db.yaml` requires `openmrs-backend.db.username` to contain `openmrs_<tenant>`, but three places here still teach `<tenant>_user`: the install command at line 153, this sentence, and the `coast` example at line 187.

I ran the `coast` block exactly as written:

```
$ helm template coast helm/openmrs-tenant \
    --set global.tenant.name=coast --set global.defaultStorageClass=standard \
    --set openmrs-backend.db.url="jdbc:mariadb:loadbalance://openmrs-mariadb.openmrs.svc.cluster.local:3306/openmrs_coast?autoReconnect=true&..." \
    --set openmrs-backend.db.hostname=openmrs-mariadb.openmrs.svc.cluster.local \
    --set openmrs-backend.db.port=3306 \
    --set openmrs-backend.db.username=coast_user \
    --set openmrs-backend.db.password=CoastPass123
Error: execution error at (openmrs-tenant/templates/validate-db.yaml:25:4): openmrs-tenant: openmrs-backend.db.username (coast_user) must contain the tenant name (openmrs_coast). This prevents one tenant from taking over another's database user. Use a tenant-scoped name (e.g. openmrs_coast_user) or set dbBootstrap.enabled=false.
```

If merged as-is, anyone copying the install command hits that on `helm install`, and a tenant that was created on the old instructions carries `<tenant>_user` in its values, so `helm upgrade` onto this chart stops the same way until they either rename a live MariaDB account or set `dbBootstrap.enabled=false`. The same command renders clean on main.

Renaming the examples to `openmrs_<tenant>_user` / `openmrs_coast_user` is the smaller fix and leaves the guard as strict as it is, so that's the one I'd take. Two more spots want a look while you're in here: the values table row for `openmrs-backend.db.username` (line 244) is where people go for the rule and says nothing about it, and the manual-fallback SQL at lines 128-129 still creates `<tenant>_user`, which is fine under `dbBootstrap.enabled=false` but means a later flip back to the default lands on this same error.

The other way out is to check the tenant name instead of the `openmrs_`-prefixed database name (`contains (.Values.global.tenant.name | toString | replace "-" "_") ($db.username | toString)`), which keeps `coast_user` legal and still rejects the primary's `openmrs`. I'd rather not: the current form is the harder one to satisfy by accident, and it's the docs that are behind here, not the guard.

```suggestion
`dbBootstrap.database`), the `openmrs_<tenant>_user` user, and its grants — the password is
```

.github/workflows/helm-build.yaml
```diff
@@ -144,19 +144,55 @@ jobs:
...
           fi
-          # DB connection is required: a tenant without db.url/hostname/username/password
-          # must fail to render, not silently fall back to the shared backend's
-          # primary-DB defaults (host "mariadb"/db "openmrs").
+          # Phase 2: operator Database/User/Grant CRs + credential Secret render,
+          # DB defaults to openmrs_<tenant>.
+          grep -q 'kind: Database' /tmp/tenant.yaml
+          grep -q 'kind: User' /tmp/tenant.yaml
+          grep -q 'kind: Grant' /tmp/tenant.yaml
+          grep -q 'name: tenant1-db-credentials' /tmp/tenant.yaml
```
dkayiwa, line 162: This block pins `cleanupPolicy: Skip` and `maxUserConnections: 20` but not `k8s.mariadb.com/watch`, and that label is the one here whose loss is silent. Drop it and the operator stops watching the credentials Secret, so `helm upgrade --set openmrs-backend.db.password=<new>` rewrites both Secrets, the backend rolls on its `checksum/secret` and comes back on a password MariaDB doesn't hold yet, and the tenant sits in CrashLoopBackOff until the periodic SQL requeue fires (`--requeue-sql` defaults to 10h, plus up to 1h of offset), with `helm upgrade` reporting success the whole time. The two values it does pin cost a lot less when they drift.

I checked the rendered Secret carries the label verbatim, so a plain grep is enough:

```suggestion
          grep -q 'name: tenant1-db-credentials' /tmp/tenant.yaml
          grep -q 'k8s.mariadb.com/watch: ""' /tmp/tenant.yaml
```

helm/openmrs-tenant/templates/db-bootstrap-grant.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: {{ .Release.Name }}-grant
+  labels:
+    {{- include "openmrs-tenant.labels" . | nindent 4 }}
+spec:
+  mariaDbRef:
+    name: {{ .Values.dbBootstrap.mariaDbRef.name | quote }}
+    namespace: {{ .Values.dbBootstrap.mariaDbRef.namespace | quote }}
+  privileges:
+    - "ALL PRIVILEGES"
+  database: {{ include "openmrs-tenant.dbName" . | quote }}
```
dkayiwa: `_` is a wildcard in a GRANT database name, so this hands a hyphenated tenant privileges on other tenants' databases. Backticks don't switch that off: the operator's `escapeWildcard` in `pkg/sql/sql.go` only wraps the value, so what reaches the server is ``GRANT ALL PRIVILEGES ON `openmrs_east_coast`.*``, which matches any database of the same length that agrees outside the `_` positions.

I ran it on a kind cluster (operator 26.6.0, primary `openmrs-mariadb` in `openmrs`) with tenants `east-coast` and `eastxcoast`. Both provisioned `Ready`, and then east-coast's user read and dropped a table in eastxcoast's database:

```
$ mariadb -uopenmrs_east_coast_user -pEcPass123 \
    -e "SELECT * FROM openmrs_eastxcoast.patient; DROP TABLE openmrs_eastxcoast.patient;"
id      name
1       private-patient-record

$ mariadb -uroot -e "SELECT COUNT(*) FROM information_schema.tables
                     WHERE table_schema='openmrs_eastxcoast'"
0
```

How far it reaches, so this doesn't read worse than it is: the primary `openmrs` database is never in range (every pattern here is at least 9 characters, `openmrs` is 7), and two single-word tenants can't reach each other either, since their only `_` is the literal one in the `openmrs_` prefix. It takes a tenant name with a hyphen in it, which is the case the chart goes out of its way to support.

Escaping the `_` on the Grant alone fixes it. `Database.spec.name` has to stay as it is, or you end up with a database literally called `openmrs\_east\_coast`, but the Grant's database is a pattern and MariaDB honours the escape inside backticks.

```suggestion
  database: {{ include "openmrs-tenant.dbName" . | replace "_" "\\_" | quote }}
```

I put that through the operator on the same cluster. The Grant reconciles `Ready/Created`, `status.currentPrivileges` settles on `["ALL PRIVILEGES"]` with no drift churn, the tenant can still create tables in its own database, and its `SHOW DATABASES` drops back to just `openmrs_east_coast`. The tenant CI block passes unchanged, since nothing in it greps the Grant's `database:` line, so it's worth an assertion there too or the escape will quietly get dropped again later.

Not a merge blocker on its own, since it takes two tenant names that happen to line up, but I'd rather it went in with this PR than became another entry in the operator-limitation note.

suubi-joshua: Makes sense thanks @dkayiwa applied the fix.

README.md
```diff
@@ -140,10 +150,30 @@ helm install <tenant> helm/openmrs-tenant \
   --set openmrs-backend.db.url="jdbc:mariadb:loadbalance://<primary-release>-mariadb.<primary-namespace>.svc.cluster.local:3306/openmrs_<tenant>?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8&useMysqlMetadata=true" \
   --set openmrs-backend.db.hostname=<primary-release>-mariadb.<primary-namespace>.svc.cluster.local \
   --set openmrs-backend.db.port=3306 \
-  --set openmrs-backend.db.username=<tenant>_user \
+  --set openmrs-backend.db.username=openmrs_<tenant>_user \
```
dkayiwa, line 155: The primary's release and namespace get filled in three places and this command covers two. `dbBootstrap.mariaDbRef` defaults to `openmrs-mariadb` in namespace `openmrs` (values.yaml lines 20-22), so a primary deployed anywhere else, and `OPENMRS_NS` is overridable in `helm/scripts/bootstrap.sh`, gets a URL and hostname pointing at the real MariaDB and CRs pointing at one that isn't there.

`helm install` still reports success, since nothing waits on a custom resource. The Database/User/Grant sit not-Ready carrying the ref-resolver failure the operator patches in when `MariaDBObject` comes back NotFound (`pkg/controller/sql/controller.go:86`), and the backend crashloops on access denied because the user was never created. The prose below the command does mention the default, but this is the spot where someone is substituting their own values.

```suggestion
  --set openmrs-backend.db.username=openmrs_<tenant>_user \
  --set dbBootstrap.mariaDbRef.name=<primary-release>-mariadb \
  --set dbBootstrap.mariaDbRef.namespace=<primary-namespace> \
```

The `coast` example below already matches the defaults, so it can stay as it is.

.github/workflows/helm-build.yaml
```diff
@@ -165,6 +205,82 @@ jobs:
...
+          fi
+          grep -qF 'dbBootstrap.database' <<<"$err"
+          # dbBootstrap disabled → no CRs/Secret leak.
+          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+            --set dbBootstrap.enabled=false > /tmp/tenant-manual.yaml
+          if grep -q 'db-bootstrap' /tmp/tenant-manual.yaml; then
+            echo "expected NO dbBootstrap resources when disabled"; exit 1
+          fi
+          # No imperative bootstrap: no Job, no rootPassword, no SQL/bash
+          # interpolation — the password lives only in a base64 Secret.
```
dkayiwa: This still says base64, which the Secret stopped doing: it renders `stringData` with `password: "ci-pass"` verbatim, and the backend subchart's own Secret sits a few resources above it in the same `/tmp/tenant.yaml` with `OMRS_DB_PASSWORD: "ci-pass"`. Worth not leaving the claim in the tree, since it's the sort of thing someone leans on when deciding a render is safe to paste somewhere.

```suggestion
          # No imperative bootstrap: no Job, no rootPassword, no SQL or bash
          # interpolation reaching the database.
```

.github/workflows/helm-build.yaml
```diff
@@ -165,6 +205,82 @@ jobs:
...
+              --set 'dbBootstrap.database=openmrs_tenant1\\db' 2>&1); then
+            echo "expected the guard to reject a database name with a backslash"; exit 1
+          fi
+          grep -qF 'single quotes, backslashes or backticks' <<<"$err"
+          # The URL's database segment must match the provisioned database — they are the same DB.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set "openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/other_db?autoReconnect=true" 2>&1); then
+            echo "expected the guard to reject a db.url whose database differs from dbBootstrap's"; exit 1
+          fi
+          grep -qF 'openmrs-tenant: openmrs-backend.db.url must target the same database' <<<"$err"
```
dkayiwa, line 297: Nothing in this step pins the order of the two splits in the drift guard. I put `first (splitList "?" (last (splitList "/" $db.url)))` back in place of the current form and the whole step stayed green, so the `serverSslCert` reading can come back without CI noticing. One render that fails under the old order covers it:

```suggestion
          grep -qF 'openmrs-tenant: openmrs-backend.db.url must target the same database' <<<"$err"
          # A '/' inside the JDBC properties must not be read as the database segment.
          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
            --set "openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/openmrs_tenant1?useSsl=true&serverSslCert=/etc/ssl/certs/mariadb-ca.pem" > /dev/null
```

I ran it both ways: it passes on the current template and fails with `the URL's database segment is mariadb-ca.pem` on the reordered one.

helm/openmrs-tenant/templates/validate-db.yaml
dkayiwa: The guard is a substring test, so this is still reachable when one tenant's normalized name is a prefix of another's, which is exactly the overlap the `-` to `_` normalization creates. Tenant `east` renders clean holding `east-coast`'s account:

```
$ helm template east helm/openmrs-tenant \
    --set global.tenant.name=east \
    --set openmrs-backend.db.url='jdbc:mariadb:loadbalance://mariadb.example:3306/openmrs_east?autoReconnect=true' \
    --set openmrs-backend.db.hostname=mariadb.example \
    --set openmrs-backend.db.username=openmrs_east_coast_user \
    --set openmrs-backend.db.password=ecpass
...
kind: User
  name: "openmrs_east_coast_user"
```

`contains "openmrs_east" "openmrs_east_coast_user"` is true, so nothing stops it, and from there it's the mechanism you traced above: `UserExists` finds the account, the operator issues `ALTER USER ... IDENTIFIED BY`, east-coast's backend is locked out on the password it still holds, and the shared account ends up carrying east's grant alongside east-coast's, so east reads and writes east-coast's schema. If merged as-is, installing `east` by copying `east-coast`'s command and changing the tenant name and the URL's database, which is the copy-paste this guard exists to catch, still does that with all three CRs reporting Ready. It catches only one direction: `openmrs_east_user` under tenant `east-coast` is rejected, the reverse is not.

An exact match on the username closes it. It has to sit below the `'`/`\` guard, or the two username cases in the workflow lose their message:

```
{{- if not (or (eq ($db.username | toString) $normalizedTenant) (eq ($db.username | toString) (printf "%s_user" $normalizedTenant))) -}}
{{- fail (printf "openmrs-tenant: openmrs-backend.db.username (%s) must be exactly %s or %s_user. Anything looser lets a tenant whose name is a prefix of another's take over that tenant's MariaDB account. Use the tenant-scoped name or set dbBootstrap.enabled=false." ($db.username | toString) $normalizedTenant $normalizedTenant) -}}
{{- end -}}
```

I ran that in place of lines 24-26, moved below the quote guard. `openmrs_east_coast_user` under tenant `east` is now rejected, and the whole `Lint and Render Tenant Chart` step passes unchanged on Helm 3.17.4, including the `openmrs_tenant1_user` / `openmrs_east_coast_user` / `openmrs_12345_user` fixtures and both `--set openmrs-backend.db.username=...'` and `...\` cases, so there's no fixture churn.

The database half I'd leave alone and stop claiming. `dbBootstrap.database` exists to name a database the tenant name doesn't derive, so no substring rule can separate "adopt my own" from "adopt east-coast's", and `--set dbBootstrap.database=openmrs_east_coast` with a matching URL does render under tenant `east` (with `cleanupPolicy=Delete` that drops east-coast's schema on uninstall, as you showed). That one takes three deliberate settings rather than a slip, so I'd just drop "This prevents one tenant from taking over another's database" from the line 22 message rather than harden it. One thing worth knowing while deciding how much that override is worth: it's install-time only either way, since `Database.spec.name` and `Grant.spec.database` are both tagged `webhook:"inmutable"` in the operator (`api/v1alpha1/database_types.go`, `grant_types.go`) and `mariadb-operator` 26.6.0 ships with `webhook.enabled: true`, so changing it on an installed release is rejected rather than applied.

.github/workflows/helm-build.yaml
suubi-joshua: I added the `serverSslCert` test at  lines 199-201, immediately after the drift guard error check at lines 193-198 same logical location you suggested, just after the existing drift guard assertions.

README.md
suubi-joshua: Added `dbBootstrap.mariaDbRef.name` and `.namespace` to the install command at `README.md` lines 155-156, right after the password line which keeps all `--set values` together in the same block rather than scattering them elsewhere in the docs.

.github/workflows/helm-build.yaml
```diff
@@ -165,6 +208,82 @@ jobs:
...
+              --set openmrs-backend.db.username=openmrs 2>&1); then
+            echo "expected the guard to reject a username that does not contain the tenant name"; exit 1
+          fi
+          grep -qF 'openmrs-backend.db.username' <<<"$err"
+          # Tenant-scoping guard: database must contain the tenant name.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set dbBootstrap.database=openmrs 2>&1); then
+            echo "expected the guard to reject a database name that does not contain the tenant name"; exit 1
+          fi
+          grep -qF 'dbBootstrap.database' <<<"$err"
```
dkayiwa: Neither of these two cases pins the guard it names, so both tenant-scoping guards can go back out without CI noticing.

`--set openmrs-backend.db.username=openmrs` was already rejected by the substring form this round replaced, so putting `contains $normalizedTenant ($db.username | toString)` back in place of the exact match on `validate-db.yaml` line 27 leaves the whole step green, and tenant `east` holding `openmrs_east_coast_user` renders again. The database case is looser still: `--set dbBootstrap.database=openmrs` breaks the URL match too, so it is the drift guard that actually fails the render, and that message ends `Fix the JDBC URL or dbBootstrap.database.`, which is enough for the `grep -qF 'dbBootstrap.database'` on line 222. I deleted lines 21-23 of `validate-db.yaml` outright and the step still passed.

Lining the URL up with the override in the database case, and adding the prefix case for the username, covers both:

```suggestion
          # Tenant-scoping guard: username must be exactly openmrs_<tenant> or openmrs_<tenant>_user.
          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
              --set openmrs-backend.db.username=openmrs 2>&1); then
            echo "expected the guard to reject a username that does not contain the tenant name"; exit 1
          fi
          grep -qF 'openmrs-backend.db.username' <<<"$err"
          # A tenant whose name is a prefix of another's must not reach that tenant's account.
          if err=$(helm template east helm/openmrs-tenant \
              --set global.tenant.name=east \
              --set openmrs-backend.db.url=jdbc:loadbalance://mariadb.example:3306/openmrs_east \
              --set openmrs-backend.db.hostname=mariadb.example \
              --set openmrs-backend.db.username=openmrs_east_coast_user \
              --set openmrs-backend.db.password=ec-pass 2>&1); then
            echo "expected the guard to reject another tenant's user"; exit 1
          fi
          grep -qF 'must be exactly openmrs_east or openmrs_east_user' <<<"$err"
          # Tenant-scoping guard: database must contain the tenant name. The URL agrees with
          # the override here, so only the tenant-scoping guard can be what rejects it.
          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
              --set dbBootstrap.database=openmrs \
              --set "openmrs-backend.db.url=jdbc:mariadb:loadbalance://mariadb.example.svc.cluster.local:3306/openmrs?autoReconnect=true" 2>&1); then
            echo "expected the guard to reject a database name that does not contain the tenant name"; exit 1
          fi
          grep -qF 'dbBootstrap.database (openmrs) must contain the tenant name' <<<"$err"
```

I ran the three of them on Helm 3.17.4. They pass on the templates as they stand; the new username block fails with `expected the guard to reject another tenant's user` once the exact match goes back to `contains`; and the database block fails with `expected the guard to reject a database name that does not contain the tenant name` once lines 21-23 are gone. Nothing here changes what the chart does today so it isn't a blocker, but the guard it covers is the one that went in this round.

README.md
```diff
@@ -208,8 +243,14 @@ through. For the full shared-chart surface, see `helm/openmrs-backend/values.yam
 | `openmrs-backend.db.url` | Full JDBC URL to the shared MariaDB | **required** |
 | `openmrs-backend.db.hostname` | Shared MariaDB host, used by the backend's `wait-for-it` preflight (`OMRS_DB_HOSTNAME`) | **required** |
 | `openmrs-backend.db.port` | Shared MariaDB port | `3306` |
-| `openmrs-backend.db.username` | Tenant DB user | **required** |
+| `openmrs-backend.db.username` | Tenant DB user (must contain `openmrs_<tenant>` when `dbBootstrap.enabled=true`) | **required** |
```
dkayiwa: This row now states a rule the chart rejects, and it is worth fixing before this merges. The username guard moved from a substring test to an exact match, so "contain" holds only for the two names it accepts:

```
$ helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
    --set openmrs-backend.db.username=openmrs_tenant1_db_user
Error: execution error at (openmrs-tenant/templates/validate-db.yaml:28:4): openmrs-tenant: openmrs-backend.db.username (openmrs_tenant1_db_user) must be exactly openmrs_tenant1 or openmrs_tenant1_user. ...
```

This table is where people come for the rule, so if merged as-is anyone who reads it and picks a tenant-scoped name of their own gets that failure on `helm install`, on a value the docs told them was fine.

```suggestion
| `openmrs-backend.db.username` | Tenant DB user; must be exactly `openmrs_<tenant>` or `openmrs_<tenant>_user` when `dbBootstrap.enabled=true` | **required** |
```

The `dbBootstrap.database` row five lines down has the same gap the other way round: it documents the URL match but not that the name has to contain `openmrs_<tenant>`.

helm/openmrs-tenant/templates/validate-db.yaml
dkayiwa: The database half of the line 22 message is still making the claim. `--set dbBootstrap.database=openmrs_east_coast` with a URL to match renders clean under tenant `east` and produces a `Database` CR named `openmrs_east_coast`, `cleanupPolicy` and all, so "This prevents one tenant from taking over another's database." is now the one sentence in the guard set that isn't true. Dropping it from the printf leaves the rest of the message doing its job.

suubi-joshua:
@dkayiwa I have addressed your review concerns

README.md
```diff
@@ -110,16 +110,26 @@ configuration on top of them. Each tenant is a separate Helm release in its own
...
+> rendered and you must create the database and user yourself, e.g.:
+>
+> ```bash
+> kubectl exec -n openmrs svc/openmrs-mariadb -- mysql -u root -pRoot123 -e "
+>   CREATE DATABASE IF NOT EXISTS openmrs_<tenant> CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
+>   CREATE USER IF NOT EXISTS 'openmrs_<tenant>_user'@'%' IDENTIFIED BY '<password>';
+>   GRANT ALL PRIVILEGES ON openmrs_<tenant>.* TO 'openmrs_<tenant>_user'@'%';
+>   FLUSH PRIVILEGES;
+> "
+> ```
```
dkayiwa, line 134: This block hands out the same wildcard grant that `db-bootstrap-grant.yaml:14` now escapes, and it is worth fixing before this merges. `_` is a pattern character in a GRANT database name, and the hyphen note further down tells people to write tenant `east-coast` as `openmrs_east_coast`, so the two underscores make this a pattern matching any same-length database that agrees outside those positions.

I ran the SQL from this block as written against MariaDB 11.4, for tenants `east-coast` and `eastxcoast`:

```
$ mariadb -uopenmrs_east_coast_user -pEcPass123 \
    -e "SELECT * FROM openmrs_eastxcoast.patient; DROP TABLE openmrs_eastxcoast.patient;"
id      name
1       private-patient-record

$ mariadb -uroot -e "SELECT COUNT(*) FROM information_schema.tables
                     WHERE table_schema='openmrs_eastxcoast'"
0
```

If merged as-is, anyone taking the documented `dbBootstrap.enabled=false` route for a hyphenated tenant gets read, write and DROP on another tenant's schema, and no chart guard can catch it, because the chart is not in that path.

The escape needs backticks around the name, and backticks inside the current `-e "..."` get command-substituted by the shell before mysql ever sees them. A quoted heredoc avoids that and keeps the SQL people read identical to the SQL that runs:

````suggestion
> ```bash
> kubectl exec -i -n openmrs svc/openmrs-mariadb -- mysql -u root -pRoot123 <<'SQL'
> CREATE DATABASE IF NOT EXISTS openmrs_<tenant> CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> CREATE USER IF NOT EXISTS 'openmrs_<tenant>_user'@'%' IDENTIFIED BY '<password>';
> GRANT ALL PRIVILEGES ON `openmrs\_<tenant>`.* TO 'openmrs_<tenant>_user'@'%';
> FLUSH PRIVILEGES;
> SQL
> ```
````

Every `_` in the name needs the escape, so `east-coast` is `` `openmrs\_east\_coast` `` and not `` `openmrs\_east_coast` ``, which is probably worth a sentence next to the block since the placeholder hides it. I ran the heredoc version on the same server: the cross-tenant SELECT comes back `ERROR 1142`, the tenant's `SHOW DATABASES` drops back to its own database, and it can still create and write tables in `openmrs_east_coast`.

File: helm/openmrs-tenant/values.yaml [hunk 18]
```diff
@@ -11,6 +11,25 @@ global:
...
+# leaves the operator. Disable for manual CREATE DATABASE/USER/GRANT.
+dbBootstrap:
+  enabled: true
+  # The shared MariaDB CR to provision on (from the primary openmrs chart).
+  mariaDbRef:
+    name: openmrs-mariadb
+    namespace: openmrs
+  # Database name override; default openmrs_<tenant> (hyphens → underscores).
+  # Must be the database openmrs-backend.db.url connects to — the chart rejects
+  # a mismatch at render time.
```
dkayiwa: These three lines describe only the URL match, but the guard also requires the name to contain `openmrs_<tenant>`, so someone reading `helm show values` picks a tenant-scoped name of their own and gets a render failure on a value this comment implies is fine.

```suggestion
  # Database name override; default openmrs_<tenant> (hyphens → underscores).
  # Must contain openmrs_<tenant> and be the database openmrs-backend.db.url
  # connects to — the chart rejects either at render time.
```

`openmrs-backend.db.username` on line 49 carries the stricter rule (exactly `openmrs_<tenant>` or `openmrs_<tenant>_user`) with no comment at all, so it is worth a line too while you are in here. Small either way, not worth a round on its own.

helm/openmrs-tenant/templates/validate-db.yaml
dkayiwa: One more way in, and the exact match cannot see this one: `-` and `_` both normalize to `_`, so two different tenant names land on the same account. Tenants `east-coast` and `east_coast` render identical CR fields on Helm 3.17.4:

```
$ helm template a helm/openmrs-tenant --set global.tenant.name=east-coast ...
  name: "openmrs_east_coast"
  database: "openmrs\\_east\\_coast"
  username: "openmrs_east_coast_user"
  name: "openmrs_east_coast_user"

$ helm template a helm/openmrs-tenant --set global.tenant.name=east_coast ...
  name: "openmrs_east_coast"
  database: "openmrs\\_east\\_coast"
  username: "openmrs_east_coast_user"
  name: "openmrs_east_coast_user"
```

Both pass every guard, because `openmrs_east_coast_user` is the exact expected name for either tenant. From there it is the path you traced: the second install's `UserExists` comes back true, so `internal/controller/user_controller.go` takes the `AlterUser` branch at line 153 instead of the create path, the first tenant's backend is locked out on the password it still holds, and both tenants carry `ALL PRIVILEGES` on the one database. I reproduced the SQL half on MariaDB 11.4, and after `ALTER USER ... IDENTIFIED BY` the first password gives `ERROR 1045 (28000): Access denied`.

Making the normalization injective closes it, and nothing in the workflow uses `_` in a tenant name so there is no fixture churn:

```
{{- if contains "_" (.Values.global.tenant.name | toString) -}}
{{- fail (printf "openmrs-tenant: global.tenant.name (%s) must not contain underscores while dbBootstrap.enabled=true, since '-' is normalized to '_' for the database name and the two forms would share one MariaDB account. Use '-' instead, or set dbBootstrap.enabled=false." (.Values.global.tenant.name | toString)) -}}
{{- end -}}
```

I put that just below line 20 and the whole `Lint and Render Tenant Chart` step passes unchanged, with `east_coast` rejected and `east-coast` still rendering as before. Not a blocker, since it takes two tenants whose names differ only in `-` versus `_`, but it is the last gap left in this guard.

README.md
suubi-joshua: My bad forgot to fix this. Though I have used a concrete openmrs_east_coast example with all _ escaped because escaping only the first _ (as in openmrs\_<tenant>) still leaks.

.github/workflows/helm-build.yaml
```diff
@@ -144,19 +144,62 @@ jobs:
...
+          grep -q 'kind: User' /tmp/tenant.yaml
+          grep -q 'kind: Grant' /tmp/tenant.yaml
+          grep -q 'name: tenant1-db-credentials' /tmp/tenant.yaml
+          grep -q 'k8s.mariadb.com/watch: ""' /tmp/tenant.yaml
+          grep -q 'mariaDbRef:' /tmp/tenant.yaml
+          grep -q 'name: "openmrs-mariadb"' /tmp/tenant.yaml
+          grep -q 'namespace: "openmrs"' /tmp/tenant.yaml
+          grep -q 'name: "openmrs_tenant1"' /tmp/tenant.yaml
+          grep -q 'maxUserConnections: 20' /tmp/tenant.yaml
+          grep -q 'cleanupPolicy: "Skip"' /tmp/tenant.yaml
```
dkayiwa: Worth pinning `characterSet` and `collate` here too. They are the two fields in these CRs whose loss is silent: the CRD defaults them (`characterSet: utf8`, `collate: utf8_general_ci` in mariadb-operator-crds 26.6.0), so dropping the two lines from `db-bootstrap-database.yaml` still gives a Ready `Database`, just a 3-byte one that diverges from the primary's `character_set_server=utf8mb4`, and it surfaces later as `Incorrect string value` on the first 4-byte character a tenant stores. Everything else in the three CRs either fails loudly when it goes (`mariaDbRef`, `privileges`, `username`) or is already asserted in this block.

I deleted both lines from the template and the whole `Lint and Render Tenant Chart` step stayed green on Helm 3.17.4. With these two greps it fails on that mutation and passes unchanged on the templates as they stand.

```suggestion
          grep -q 'cleanupPolicy: "Skip"' /tmp/tenant.yaml
          grep -q 'characterSet: utf8mb4' /tmp/tenant.yaml
          grep -q 'collate: utf8mb4_general_ci' /tmp/tenant.yaml
```

helm/openmrs-tenant/templates/validate-db.yaml
dkayiwa: One more pair the exact match still lets through, so the underscore guard was not quite the last gap: the two accepted forms overlap whenever one tenant's name is another's plus `-user`. Tenants `east` and `east-user` both render a `User` CR named `openmrs_east_user` on Helm 3.17.4:

```
$ helm template east helm/openmrs-tenant --set global.tenant.name=east \
    --set openmrs-backend.db.url=jdbc:loadbalance://mariadb.example:3306/openmrs_east \
    --set openmrs-backend.db.hostname=mariadb.example \
    --set openmrs-backend.db.username=openmrs_east_user --set openmrs-backend.db.password=p1
...
kind: User
  name: "openmrs_east_user"

$ helm template eastuser helm/openmrs-tenant --set global.tenant.name=east-user \
    --set openmrs-backend.db.url=jdbc:loadbalance://mariadb.example:3306/openmrs_east_user \
    --set openmrs-backend.db.hostname=mariadb.example \
    --set openmrs-backend.db.username=openmrs_east_user --set openmrs-backend.db.password=p2
...
kind: User
  name: "openmrs_east_user"
```

`openmrs_east_user` is the `_user` form for tenant `east` and the bare form for tenant `east-user`, so both render clean, and from there it is the path we traced before: the second install's `UserExists` comes back true, the operator takes the `AlterUser` branch, `east`'s backend is locked out on the password it still holds, and the one account ends up carrying `ALL PRIVILEGES` on both `openmrs\_east` and `openmrs\_east\_user`.

Dropping the bare `openmrs_<tenant>` form closes it outright rather than one case at a time. With `_` now rejected in tenant names the normalization is injective, so a single fixed suffix makes tenant to username injective too, and `B = A + "-user"` is the only overlap the two-form rule can produce:

```
{{- if ne ($db.username | toString) (printf "%s_user" $normalizedTenant) -}}
{{- fail (printf "openmrs-tenant: openmrs-backend.db.username (%s) must be exactly %s_user. Anything looser lets one tenant take over another tenant's MariaDB account. Use the tenant-scoped name or set dbBootstrap.enabled=false." ($db.username | toString) $normalizedTenant) -}}
{{- end -}}
```

Nothing uses the bare form today, so there is almost no churn. I ran that in place of lines 30-32 and the whole `Lint and Render Tenant Chart` step passes on Helm 3.17.4 with one string changed: `grep -qF 'must be exactly openmrs_east or openmrs_east_user'` at helm-build.yaml:233 becomes `must be exactly openmrs_east_user`. `east-user` is then rejected, `east` keeps `openmrs_east_user`, and the `openmrs` and `openmrs_east_coast_user` negative cases behave as before. The values-table row at README.md:250 wants the same trim.

Not a blocker, since it takes a tenant literally named `<something>-user`, but it is cheaper to close than to leave as the one case this guard does not cover.

.github/workflows/helm-build.yaml
suubi-joshua: Confirmed this, I added `grep -q 'characterSet: utf8mb4'` and `grep -q 'collate: utf8mb4_general_ci'` right after the cleanupPolicy check as you suggested and this fixes.

.github/workflows/helm-build.yaml
```diff
@@ -144,19 +144,64 @@ jobs:
...
```
dkayiwa: Worth making this one a count instead of a plain grep. It matches any of the three CRs, so `cleanupPolicy` can go missing from `db-bootstrap-database.yaml` and this step stays green, and that is the one field in these CRs whose loss destroys data rather than degrading it: the operator treats a nil `cleanupPolicy` as `Delete` (`ptr.Deref(resource.CleanupPolicy(), CleanupPolicyDelete)` in `pkg/controller/sql/finalizer.go`), so `helm uninstall <tenant>` would take the tenant's schema with it.

I ran both halves on a kind cluster (operator 26.6.0, primary `openmrs-mariadb` in `openmrs`, tenants in their own namespaces). With the line as the chart renders it, deleting the `Database` CR leaves the data alone:

```
$ kubectl -n tenant-eastxcoast delete database eastxcoast-db
database.k8s.mariadb.com "eastxcoast-db" deleted
$ mariadb -uroot -e "SELECT * FROM openmrs_eastxcoast.keepme"
7
```

With the line dropped from the template, the CR still reports `Ready`, and deleting it drops the schema:

```
$ kubectl -n tenant-nocleanup get database nocleanup-db \
    -o custom-columns=READY:.status.conditions[0].status,CLEANUP:.spec.cleanupPolicy
READY   CLEANUP
True    <none>
$ kubectl -n tenant-nocleanup delete database nocleanup-db
$ mariadb -uroot -e "SELECT COUNT(*) FROM information_schema.SCHEMATA
                     WHERE SCHEMA_NAME='openmrs_nocleanup'"
0
```

A count pins all three CRs at once:

```suggestion
          [ "$(grep -c 'cleanupPolicy: "Skip"' /tmp/tenant.yaml)" -eq 3 ]
```

I ran that on Helm 3.17.4: the whole `Lint and Render Tenant Chart` step passes unchanged on the templates as they stand, and fails on the mutation above. Nothing is wrong in the tree today, so this isn't a blocker. `namespace: "openmrs"` on line 166 has the same shape (three matches, any one of them satisfies it), but losing that one fails loudly since the CRs never reconcile, so I'd leave it as it is.

suubi-joshua: Seems like a minor fix, I have added anyway to just harden things up.

helm/openmrs-tenant/values.yaml
```diff
@@ -11,6 +11,25 @@ global:
   mariadb:
     enabled: false
 
+# dbBootstrap provisions the tenant database/user/grants on the shared MariaDB
+# via the MariaDB operator's Database/User/Grant CRs — the root password never
+# leaves the operator. Disable for manual CREATE DATABASE/USER/GRANT.
+dbBootstrap:
+  enabled: true
```
dkayiwa, line 18: Worth bumping `helm/openmrs-tenant/Chart.yaml` in this commit too. Turning this on by default makes the guards apply to every tenant, and three value shapes that render on `main` today are rejected here:

* `--set openmrs-backend.db.username=coast_user` for tenant `coast`, now `validate-db.yaml:31` (must be exactly `openmrs_coast_user`)
* `--set global.tenant.name=east_coast`, now `validate-db.yaml:20` (no underscores in the tenant name)
* a URL ending `/coastdb` for tenant `coast`, now `validate-db.yaml:42` (must target `openmrs_coast`)

I ran all three against `origin/main` and against this branch on Helm 3.17.4. Each renders clean on main and fails here, so this changes the chart's value contract rather than adding to it, and the version is the only thing a consumer sees.

`README.md:601` and the `Release Charts` workflow input both say this chart versions independently of the umbrella and gets bumped by hand in a regular commit, and #30 did that for `openmrs-backend` / `openmrs-frontend` (2.0.0 to 2.1.0) in the same commit that changed their templates. Left at 1.0.0 the published chart won't tell the dbBootstrap version apart from the one without it. I'd make it 2.0.0 for what breaks above.

helm/openmrs-tenant/templates/validate-db.yaml
```diff
@@ -10,3 +9,36 @@ the shared backend falls back to its own defaults (host "mariadb", db "openmrs"
...
+Why the guard: the MariaDB operator (26.6.0) builds SQL from CR fields without
+escaping ' or \, so those chars would leave a User/Grant CR stuck in Error.
+CRD presence can't be checked at render time (helm template runs without a
+cluster), so a missing operator is a documented README prerequisite.
+*/}}
+{{- if contains "_" (.Values.global.tenant.name | toString) -}}
+{{- fail "openmrs-tenant: global.tenant.name must not contain underscores while dbBootstrap.enabled=true — both - and _ normalize to _ in the database/user names, so tenants east-coast and east_coast would collide on the same MariaDB account. Use hyphens only or set dbBootstrap.enabled=false." -}}
+{{- end -}}
+{{- $dbName := include "openmrs-tenant.dbName" . -}}
+{{- $normalizedTenant := (printf "openmrs_%s" (.Values.global.tenant.name | toString | replace "-" "_")) -}}
```
dkayiwa: This repeats the expression `openmrs-tenant.dbName` uses for its own default, so the `-` to `_` normalization now lives in two files that have to move together. Whoever changes it next has to remember both; miss one and the guard on line 30 measures the username against one normalization while the CRs provision a database from the other, so the render fails on any tenant name where the two disagree with a message that points at neither file.

`_helpers.tpl` is where this PR already put the shared part, so pulling the inner half out finishes it:

```gotemplate
{{- /*
The tenant-scoped name openmrs_<tenant> (hyphens become underscores so it stays
consistent with the openmrs_<tenant> JDBC URL convention). Both the default
database name and the required DB username are built from it.
*/}}
{{- define "openmrs-tenant.normalizedTenant" -}}
{{- printf "openmrs_%s" (.Values.global.tenant.name | toString | replace "-" "_") -}}
{{- end }}

{{- define "openmrs-tenant.dbName" -}}
{{- .Values.dbBootstrap.database | default (include "openmrs-tenant.normalizedTenant" .) -}}
{{- end }}
```

and this line becomes:

```suggestion
{{- $normalizedTenant := include "openmrs-tenant.normalizedTenant" . -}}
```

I ran that on Helm 3.17.4: the whole `Lint and Render Tenant Chart` step passes unchanged, and the renders for `east-coast` and for a `dbBootstrap.database` override come out byte-identical to the current output. Nothing is wrong in the tree today, so this is only a cleanup.

helm/openmrs-tenant/values.yaml
dkayiwa: The `openmrs-backend.db.username` half of this is still open. Line 49 has no comment, and the rule tightened after I wrote the above (it is exactly `openmrs_<tenant>_user` now, the bare `openmrs_<tenant>` form is gone), so `helm show values` shows the strictest required value in the chart with nothing saying what it has to be.

```yaml
    # Must be exactly openmrs_<tenant> + "_user" (hyphens → underscores) while
    # dbBootstrap.enabled=true; the chart rejects anything else at render time.
    username: ""
```

dkayiwa (approved):

helm/openmrs-tenant/values.yaml
suubi-joshua: I agree, bumped version

PR #32, opened 2026-08-24
TRUNK-6654: Add per-tenant host-scoped HTTPRoute routing via Gateway API.

suubi-joshua:
## Summary
Previously, all tenants and the primary stack shared the same hostless HTTPRoutes on the Traefik Gateway, meaning there was no way to route traffic to a specific tenant based on hostname and every request hit the primary catch-all. This PR adds `gateway.hostnames` to the shared backend and frontend charts so that tenant HTTPRoutes emit host-scoped hostname matches while the primary umbrella's routes remain hostless as a catch-all. A validation guard in the tenant chart prevents tenants from enabling routing without setting hostnames, which would collide on the shared gateway. The result is that `kampala.example.com/openmrs/spa` routes to the kampala tenant, `masaka.example.com/openmrs/spa` routes to masaka, and any unmatched host falls through to primary, all on a single shared Traefik Gateway with zero manual routing configuration.

JIRA TICKET: https://openmrs.atlassian.net/browse/TRUNK-6654

### DIFF
--- a/.github/workflows/helm-build.yaml
+++ b/.github/workflows/helm-build.yaml
@@ -149,10 +149,10 @@ jobs:
           if grep -q 'storageClassName: ""' /tmp/tenant.yaml; then
             echo "empty storageClassName means no-class (PVC stays Pending); it must be omitted"; exit 1
           fi
-          # Phase-1 invariant: HTTPRoutes stay off (host-less routes would collide
-          # on the shared gateway) until Phase 3 (TRUNK-6654).
+          # Default render: HTTPRoutes stay off (gateway.enabled=false by default
+          # in the tenant chart, so hostless routes would collide on the shared gateway).
           if grep -q 'kind: HTTPRoute' /tmp/tenant.yaml; then
-            echo "expected NO HTTPRoutes in the default tenant render (gateway must stay off until Phase 3)"; exit 1
+            echo "expected NO HTTPRoutes in the default tenant render (gateway must stay off by default)"; exit 1
           fi
           # Phase 2: operator Database/User/Grant CRs + credential Secret render,
           # DB defaults to openmrs_<tenant>.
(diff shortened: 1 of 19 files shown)


suubi-joshua:
@claude review

helm/openmrs-tenant/templates/validate.yaml
```diff
@@ -42,3 +42,12 @@ cluster), so a missing operator is a documented README prerequisite.
...
 {{- end -}}
+{{- /* Phase 3: routing guard — host-scoped HTTPRoutes must not collide. */ -}}
+{{- $beGw := (index .Values "openmrs-backend").gateway | default dict -}}
+{{- $feGw := (index .Values "openmrs-frontend").gateway | default dict -}}
+{{- if and (dig "enabled" false $beGw) (not (dig "hostnames" nil $beGw)) -}}
+{{- fail "openmrs-tenant: openmrs-backend.gateway.enabled=true requires openmrs-backend.gateway.hostnames to be set — a hostless HTTPRoute would collide on the shared gateway." -}}
+{{- end -}}
+{{- if and (dig "enabled" false $feGw) (not (dig "hostnames" nil $feGw)) -}}
+{{- fail "openmrs-tenant: openmrs-frontend.gateway.enabled=true requires openmrs-frontend.gateway.hostnames to be set — a hostless HTTPRoute would collide on the shared gateway." -}}
+{{- end -}}
```
dkayiwa, line 53: This needs a guard extension before merge. The two checks here are per-route, so a tenant can turn on the frontend route without the backend one, and that combination silently points the tenant's users at the primary's database.

Rendered it to be sure: `--set openmrs-frontend.gateway.enabled=true --set "openmrs-frontend.gateway.hostnames[0]=coast.example.com"` with no backend flags renders clean and produces exactly one HTTPRoute, `coast.example.com` + `PathPrefix /openmrs/spa` to the tenant frontend. The same render sets `API_URL: "/openmrs"`, a relative path, so the SPA calls `coast.example.com/openmrs/ws/rest/v1/...`. The tenant has nothing matching that path. The primary umbrella does, through the hostless `PathPrefix /openmrs/` catch-all this whole design leans on (confirmed in the primary render). So the tenant's SPA loads on the tenant's own hostname and reads and writes the primary instance, with nothing thrown and nothing logged.

A typo lands in the same place, since the hostname is now set in two independent lists with nothing tying them together: backend on `coast.example.com`, frontend on `cosat.example.com`, and `coast.example.com/openmrs/spa` quietly serves the primary's SPA.

Extending the guard covers both. I ran this version through the cases: frontend-only and mismatched hostnames now fail, backend-only (an API-only tenant) still renders, an absolute `openmrs-frontend.apiUrl` stays an escape hatch, the enabled-without-hostnames rejections still fire, and every routing case the new CI steps exercise still passes.

```suggestion
{{- if and (dig "enabled" false $beGw) (not (dig "hostnames" nil $beGw)) -}}
{{- fail "openmrs-tenant: openmrs-backend.gateway.enabled=true requires openmrs-backend.gateway.hostnames to be set — a hostless HTTPRoute would collide on the shared gateway." -}}
{{- end -}}
{{- if and (dig "enabled" false $feGw) (not (dig "hostnames" nil $feGw)) -}}
{{- fail "openmrs-tenant: openmrs-frontend.gateway.enabled=true requires openmrs-frontend.gateway.hostnames to be set — a hostless HTTPRoute would collide on the shared gateway." -}}
{{- end -}}
{{- /* The SPA calls the API same-origin while apiUrl is a relative path, so every host that
       serves the tenant SPA must also route /openmrs/ to this tenant's backend. Otherwise
       those calls fall through the primary umbrella's hostless route to the primary DB. */ -}}
{{- $apiUrl := (index .Values "openmrs-frontend").apiUrl | default "" -}}
{{- if and (dig "enabled" false $feGw) (hasPrefix "/" $apiUrl) -}}
{{- $beHosts := ternary (dig "hostnames" list $beGw) list (dig "enabled" false $beGw) -}}
{{- range $h := (dig "hostnames" list $feGw) -}}
{{- if not (has $h $beHosts) -}}
{{- fail (printf "openmrs-tenant: %s serves the tenant SPA but openmrs-backend.gateway does not route it, so the SPA's relative apiUrl (%s) would reach the primary backend instead. Add %s to openmrs-backend.gateway.hostnames with openmrs-backend.gateway.enabled=true, or set an absolute openmrs-frontend.apiUrl." $h $apiUrl $h) -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

.github/workflows/helm-build.yaml
```diff
@@ -310,12 +310,90 @@ jobs:
...
+            --set elasticsearch.enabled=false \
+            --set seaweedfs.enabled=false > /tmp/primary.yaml
+          grep -q 'kind: HTTPRoute' /tmp/primary.yaml
+          if grep -q 'coast.example.com\|nyanza.example.com' /tmp/primary.yaml; then
+            echo "primary render must not contain tenant hostnames"; exit 1
+          fi
+          # Primary routes must be hostless (no hostnames: block).
+          if grep -A5 'kind: HTTPRoute' /tmp/primary.yaml | grep -q 'hostnames:'; then
+            echo "primary HTTPRoutes must not have hostnames (catch-all)"; exit 1
+          fi
```
dkayiwa: This check can't fail, so I'd fix it before merging. `hostnames:` sits 14 lines below `kind: HTTPRoute` (metadata, five labels, spec, parentRefs), well outside the `-A5` window. Rendering the primary with `--set "openmrs-frontend.gateway.hostnames[0]=oops.example.com" --set "openmrs-backend.gateway.hostnames[0]=oops.example.com"` gives two routes that both carry `hostnames:`, and this step still passes. That matters because it's the only thing guarding the catch-all that every tenant's fall-through depends on, so the change that ends it would ship green.

Grepping the whole render wouldn't be right either: the primary's root-redirect route is meant to have a hostname (`localhost`, which this PR moves into `helm/openmrs/values.yaml`), so an unscoped check would start failing as soon as `rootRedirect.enabled=true`. Scoping to the two catch-all templates handles both. I checked that this version fires on the `oops.example.com` render, stays quiet on the clean one, and stays quiet with `rootRedirect.enabled=true`.

```suggestion
          # The primary backend/frontend routes must stay hostless (catch-all).
          # The root-redirect route legitimately carries a hostname, so scope this
          # to the two catch-all templates rather than the whole render.
          if helm template helm/openmrs --generate-name \
              --set monitoring.enabled=false \
              --set elasticsearch.enabled=false \
              --set seaweedfs.enabled=false \
              --show-only charts/openmrs-backend/templates/httproute.yaml \
              --show-only charts/openmrs-frontend/templates/httproute.yaml \
              | grep -q 'hostnames:'; then
            echo "primary HTTPRoutes must not have hostnames (catch-all)"; exit 1
          fi
```

helm/openmrs-tenant/templates/NOTES.txt
```diff
@@ -4,6 +4,14 @@ Tenant: {{ .Values.global.tenant.name }}
 Backend: {{ .Release.Name }}-openmrs-backend.{{ .Release.Namespace }}.svc.cluster.local:8080
 Frontend: {{ .Release.Name }}-openmrs-frontend.{{ .Release.Namespace }}.svc.cluster.local:80
 
+{{- if and (index .Values "openmrs-backend" "gateway" "enabled") (index .Values "openmrs-backend" "gateway" "hostnames") }}
+
+Tenant URL (via gateway):
+{{- range (index .Values "openmrs-backend" "gateway" "hostnames") }}
+  http://{{ . }}{{ index $.Values "openmrs-frontend" "spaPath" }}/home
+{{- end }}
```
dkayiwa: The SPA URL here is keyed on the backend route's flags, so it misreports whenever the two routes differ. From `helm install --dry-run`: backend on `be.example.com` with frontend on `fe.example.com` prints `http://be.example.com/openmrs/spa/home`, a host the tenant's SPA route doesn't claim; backend-only prints an SPA URL when no tenant SPA route exists at all; frontend-only prints nothing, which is the one case where the URL would have been correct.

Reading the frontend's own gateway values fixes all three, and the two NOTES assertions this PR adds to CI still pass with it.

```suggestion
{{- if and (index .Values "openmrs-frontend" "gateway" "enabled") (index .Values "openmrs-frontend" "gateway" "hostnames") }}

Tenant URL (via gateway):
{{- range (index .Values "openmrs-frontend" "gateway" "hostnames") }}
  http://{{ . }}{{ index $.Values "openmrs-frontend" "spaPath" }}/home
{{- end }}
```

.github/workflows/helm-build.yaml
```diff
@@ -310,12 +310,90 @@ jobs:
...
+          if grep -q '/openmrs/spa/home' helm/openmrs-tenant/templates/NOTES.txt; then
+            echo "NOTES.txt must not hardcode /openmrs/spa when spaPath is used"; exit 1
+          fi
+
+      - name: Validate backend/frontend version pins
+        run: |
+          grep -q 'version: 2.2.0' helm/openmrs/Chart.yaml
+          grep -q 'version: 2.2.0' helm/openmrs-tenant/Chart.yaml
+          grep -q 'version: 2.2.0' helm/openmrs-backend/Chart.yaml
+          grep -q 'version: 2.2.0' helm/openmrs-frontend/Chart.yaml
```
dkayiwa: I'd drop this step. `helm dependency update` at lines 40 and 43 already fails loudly on the drift it's checking for, 350 lines earlier: putting `helm/openmrs-backend/Chart.yaml` back to 2.1.0 while the pins say 2.2.0 gives `Error: can't get a valid version for 1 subchart(s): "openmrs-backend" (repository "file://../openmrs-backend", version "2.2.0")` from both `helm/openmrs` and `helm/openmrs-tenant`.

What the hardcoded `2.2.0` adds is a step that goes red on the next routine version bump, for a reason unrelated to whatever change is in flight, surfacing as a bare `grep` exit 1. It also doesn't check what its name suggests for the tenant chart: `helm/openmrs-tenant/Chart.yaml`'s own `version:` is 2.0.0, and this grep is matching its indented dependency line instead.

helm/openmrs-tenant/Chart.yaml
```diff
@@ -7,11 +7,11 @@ appVersion: "3.7.x-no-demo"
 
 dependencies:
   - name: openmrs-backend
-    version: 2.1.0
+    version: 2.2.0
```
dkayiwa, line 10: The tenant chart's own `version:` is still 2.0.0 while its templates and values changed in this PR (the routing guard, the NOTES block, `hostnames`, these pins). README's "Releasing from Github Actions" step 1 asks for `openmrs-backend`/`openmrs-frontend`/`openmrs-tenant` to be bumped by hand in the regular commit, since the release workflow only rewrites the umbrella's version, and b08ea74 did exactly that for the previous tenant-chart change. Worth bumping to 2.1.0 here.

Nothing is broken today, no release has published `openmrs-tenant` yet, so this is really about not leaving it for whoever runs the next release to spot.

helm/openmrs-tenant/values.yaml
```diff
@@ -35,6 +34,7 @@ openmrs-backend:
   # would collide on the shared gateway.
```
dkayiwa: This comment and the one at line 58 now say the opposite of what the chart does, and a reader who trusts them would conclude tenant routing still isn't available. Something like "off by default; set `hostnames` alongside `enabled`" would match. The CI job carries the same leftover at line 152 ("HTTPRoutes stay off ... until Phase 3 (TRUNK-6654)"), which is this PR.

rkorytkowski:
Also add hostnames: [localhost] (or similar) to openmrs-frontend.rootRedirect in helm/kind-openmrs.yaml, matching what was done for helm/openmrs/values.yaml.

suubi-joshua:
I have addressed all the above review concerns.

.github/workflows/helm-build.yaml
```diff
@@ -310,12 +310,121 @@ jobs:
...
+            --set openmrs-backend.db.password=ci-pass"
+          # Case 1: rootRedirect.hostnames empty, gateway.hostnames set → fallback
+          helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+            --set openmrs-backend.gateway.enabled=true \
+            --set "openmrs-backend.gateway.hostnames[0]=tenant1.example.com" \
+            --set openmrs-frontend.gateway.enabled=true \
+            --set "openmrs-frontend.gateway.hostnames[0]=tenant1.example.com" \
+            --set openmrs-frontend.rootRedirect.enabled=true > /tmp/tenant-fallback.yaml
+          # root-redirect-httproute should inherit gateway.hostnames
+          grep -A20 'root-redirect' /tmp/tenant-fallback.yaml | grep -q 'tenant1.example.com'
```
dkayiwa, line 440: Nothing in the new checks would notice if `rootRedirect.hostnames` went back to defaulting to `[localhost]` in the shared frontend chart. Not a blocker, but worth locking down while the step is being written.

Moving that default out of `helm/openmrs-frontend/values.yaml` and into `helm/openmrs/values.yaml` plus `helm/kind-openmrs.yaml` is what keeps tenants from inheriting it, and it's the one routing invariant in this PR that nothing asserts. I put the old default back and re-ran every new step on helm 3.17.4: the tenant's root-redirect renders

```
  hostnames:
    - localhost
    - tenant1.example.com
```

and all of them still pass, this one included, because `tenant1.example.com` is present and satisfies the grep. At that point the primary's route and every tenant's route all claim `Exact /` on `localhost`, and Gateway API resolves that tie by creation timestamp, so `localhost/` follows whichever route happens to be oldest. For a tenant with its own `spaPath` that means the local dev entry point redirecting into a tenant's SPA, with nothing thrown and nothing logged.

The negative assertion is cheap. I checked it stays quiet on the current render and fires once the old default is back.

```suggestion
          grep -A20 'root-redirect' /tmp/tenant-fallback.yaml | grep -q 'tenant1.example.com'
          # ...and nothing else: if the shared frontend chart ever defaults
          # rootRedirect.hostnames back to [localhost], every tenant's route would
          # claim localhost too and the check above would still pass.
          if grep -A20 'root-redirect' /tmp/tenant-fallback.yaml | grep -q 'localhost'; then
            echo "tenant root-redirect must not inherit a localhost default from the shared chart"; exit 1
          fi
```

README.md
```diff
@@ -222,16 +226,14 @@ kubectl port-forward -n tenant-<tenant> svc/<tenant>-openmrs-backend 8080:8080
...
-> tenant's frontend and backend will sit behind a per-tenant hostname
-> (e.g. `<tenant>.example.com`), like the primary OpenMRS stack.
+> **Note on routing:** tenant HTTPRoutes are disabled by default (`gateway.enabled=false`).
+> To enable host-based routing, set `openmrs-backend.gateway.enabled=true` and
+> `openmrs-backend.gateway.hostnames` (and likewise for frontend). The shared charts'
+> routes emit hostnames only when `gateway.hostnames` is non-empty, so the primary
+> umbrella's hostless routes act as a catch-all while tenant routes match on specific
+> hostnames — no collisions. Each tenant's frontend and backend sit behind a per-tenant
+> hostname (e.g. `coast.example.com`), like the primary OpenMRS stack. Access via
+> port-forward or by adding the hostname to `/etc/hosts` pointing at the gateway.
```
dkayiwa: I'd keep the port-forward caveat this rewrite dropped, though it's your call. The old note said the frontend SPA is not fully usable over port-forward, because the app shell requests its assets under `SPA_PATH` and needs the gateway's prefix rewrite to resolve them. That hasn't changed, and the chart still says so at install time: `helm/openmrs-tenant/templates/NOTES.txt` lines 34-37 print "the SPA UI is not fully usable this way yet ... (you get a blank page)".

As written the README now offers port-forward as a way in, ten lines under the `port-forward ... svc/<tenant>-openmrs-frontend 8081:80` command it shows. Someone follows it, gets the blank page, and the explanation is no longer anywhere in the README.

```suggestion
> hostname (e.g. `coast.example.com`), like the primary OpenMRS stack. Point that
> hostname at the gateway (DNS, or an `/etc/hosts` entry) to reach the SPA. Port-forward
> still works for the backend API, but not for the SPA UI: the app shell requests its
> assets under `SPA_PATH`, which needs the gateway's prefix rewrite, so you get a blank page.
```

suubi-joshua: Since I am adding hostname-based routing as the proper way to reach tenants, I had removed the broken port-forward line rather than adding a caveat. It avoids showing a command that doesn't work, and the routing note already points users to hostnames.

.github/workflows/helm-build.yaml
```diff
@@ -310,12 +310,127 @@ jobs:
...
+          # The primary backend/frontend routes must stay hostless (catch-all).
+          # The root-redirect route legitimately carries a hostname, so scope this
+          # to the two catch-all templates rather than the whole render.
+          if helm template helm/openmrs --generate-name \
+              --set monitoring.enabled=false \
+              --set elasticsearch.enabled=false \
+              --set seaweedfs.enabled=false \
+              --show-only charts/openmrs-backend/templates/httproute.yaml \
+              --show-only charts/openmrs-frontend/templates/httproute.yaml \
+              | grep -q 'hostnames:'; then
```
dkayiwa: This check still goes green when `helm template` itself fails. The pipeline's exit status is `grep`'s, and GitHub Actions runs the step under `bash -e` with no `pipefail`, so a `--show-only` path that stops resolving (a future rename of `httproute.yaml` in either chart) leaves it passing on an empty render. I pointed the backend path at a template that doesn't exist: helm prints `Error: could not find template charts/openmrs-backend/templates/renamed.yaml in chart`, grep matches nothing, step exits 0.

Nothing is broken today, so this can wait if you'd rather not reopen the step. I'd still close it, because once the primary's two routes pick up hostnames there is no other check that would notice, and unmatched hosts stop reaching the primary at all.

Capturing the output first is what the routing-guard step above already does (`err=$(...)` read back with `<<<"$err"`), and under `bash -e` the assignment aborts the step when helm errors. Checked on helm 3.17.4: the version below stays quiet on the clean primary render, still fires with `--set "openmrs-backend.gateway.hostnames[0]=oops.example.com"`, and exits 1 on the renamed-template case.

```suggestion
          out=$(helm template helm/openmrs --generate-name \
              --set monitoring.enabled=false \
              --set elasticsearch.enabled=false \
              --set seaweedfs.enabled=false \
              --show-only charts/openmrs-backend/templates/httproute.yaml \
              --show-only charts/openmrs-frontend/templates/httproute.yaml)
          if grep -q 'hostnames:' <<<"$out"; then
```

suubi-joshua: Applied the fix, since it has no harm to tighten things up.

README.md
```diff
@@ -215,23 +219,24 @@ kubectl wait --for=condition=ready pod -n tenant-<tenant> --all --timeout=600s
 # Check the MariaDB operator reconciled the tenant's CRs (Database/User/Grant ready)
 kubectl get database,user,grant -n tenant-<tenant>
 
-# Port-forward to backend (API)
-kubectl port-forward -n tenant-<tenant> svc/<tenant>-openmrs-backend 8080:8080
+# Point the tenant hostname at the gateway (local dev)
+echo "127.0.0.1 coast.example.com" | sudo tee -a /etc/hosts
+
+# Open the SPA in a browser
+# http://coast.example.com/openmrs/spa/home
```
dkayiwa: This URL needs the gateway's port before merge. The section sits under "Quick Start (Kind for local testing)", and on that cluster Traefik is only reachable on host port 8080: `helm/kind-config.yaml` maps `containerPort: 8080` to `hostPort: 8080` on `127.0.0.1`, and `helm/openmrs-operator/values.yaml` puts the web entrypoint there (`ports.web.hostPort: 8080`). So with `coast.example.com` pointed at `127.0.0.1` two lines up, `http://coast.example.com/openmrs/spa/home` reaches port 80 with nothing listening, and the browser reports connection refused.

The README already settles the convention a few sections up: the primary stack is `http://localhost:8080/openmrs/spa/login`, and the Grafana and SeaweedFS Admin rows are on `:8080` too, under "No port-forwarding needed - Traefik binds the port directly."

Worth fixing rather than leaving as a doc nit, because the frontend port-forward came out of this same block in this change, so this line is now the only documented way to open a tenant's SPA.

```suggestion
# Open the SPA in a browser (Traefik binds host port 8080 on the Kind cluster)
# http://coast.example.com:8080/openmrs/spa/home
```

helm/openmrs-tenant/templates/NOTES.txt
```diff
@@ -4,6 +4,14 @@ Tenant: {{ .Values.global.tenant.name }}
 Backend: {{ .Release.Name }}-openmrs-backend.{{ .Release.Namespace }}.svc.cluster.local:8080
 Frontend: {{ .Release.Name }}-openmrs-frontend.{{ .Release.Namespace }}.svc.cluster.local:80
 
+{{- if and (index .Values "openmrs-frontend" "gateway" "enabled") (index .Values "openmrs-frontend" "gateway" "hostnames") }}
+
+Tenant URL (via gateway):
+{{- range (index .Values "openmrs-frontend" "gateway" "hostnames") }}
+  http://{{ . }}{{ index $.Values "openmrs-frontend" "spaPath" }}/home
```
dkayiwa: The chart prints the same portless URL at install time, and this is the copy most people will actually follow, so it needs the same fix. With the README's `coast` example it renders `http://coast.example.com/openmrs/spa/home`, while the `/etc/hosts` line lower down (line 37) sends that name to `127.0.0.1`, where nothing answers on port 80.

The fix differs from the README's, because NOTES print on real clusters too and the template has no way to know the gateway's external port. I would name the local case in the heading rather than hardcode a port:

```suggestion
Tenant URL (via gateway, on the gateway's own port; :8080 on the Kind setup):
{{- range (index .Values "openmrs-frontend" "gateway" "hostnames") }}
  http://{{ . }}{{ index $.Values "openmrs-frontend" "spaPath" }}/home
```

Hardcoding `:8080` here works too if you would rather keep this block local-dev only, since the `127.0.0.1` line already assumes Kind, but then both lines should say so.

.github/workflows/helm-build.yaml
```diff
@@ -310,12 +310,129 @@ jobs:
...
+          # Mismatched hostnames: frontend on cosat.example.com, backend on coast.example.com.
+          if err=$(helm template tenant1 helm/openmrs-tenant $TENANT_VALUES \
+              --set openmrs-backend.gateway.enabled=true \
+              --set "openmrs-backend.gateway.hostnames[0]=coast.example.com" \
+              --set openmrs-frontend.gateway.enabled=true \
+              --set "openmrs-frontend.gateway.hostnames[0]=cosat.example.com" 2>&1); then
+            echo "expected the guard to reject mismatched hostnames"; exit 1
+          fi
+          grep -qF 'cosat.example.com serves the tenant SPA' <<<"$err"
+          # Backend-only (API-only tenant) must still render.
```
dkayiwa, line 377: Is an API-only tenant meant to be reachable at `/openmrs/spa` on its own hostname? On the config this line blesses it is, and what it serves is the primary's SPA.

Traefik does not give the hostname precedence over the path the way the Gateway API spec describes. In v3.3.1 (what chart 34.0.0 pins) `buildMatchRule` adds `len(hostname)` on top of the path priority instead of resolving on the hostname first, and the Gateway the operator chart renders carries no listener hostname, so `findMatchingHostnames` leaves a hostless route hostless and it contributes 0. For `tenant1.example.com/openmrs/spa/home`:

* primary frontend, hostless `/openmrs/spa`: `10000 + 12*100 + 0 + 1` = 11201
* tenant backend, `tenant1.example.com` + `/openmrs/`: `10000 + 9*100 + 19 + 1` = 10920

The primary's route wins, so the tenant's hostname serves the primary's SPA bundle and its `SPA_CONFIG_URLS`, talking back to the tenant's own backend (the tenant backend does beat the primary's for `/openmrs/ws/...`, 10920 against 10901). Nothing throws, nothing logs.

With both routes on, which is what the README documents, the tenant wins on every path and the design holds, so this is only about the asymmetric shape. If API-only is a shape you want to keep, I would add a line here recording that `/openmrs/spa` on that host falls through to the primary, so the next person does not have to redo the arithmetic. If it is not, the guard could require the frontend route whenever the backend one is on.

helm/openmrs-tenant/templates/validate.yaml
dkayiwa: The `hasPrefix "/"` test in the block above is too narrow, and that part is on me. It treats anything not starting with `/` as absolute, so a relative `apiUrl` slips through the very check that exists to catch it. On helm 3.17.4, with `openmrs-frontend.gateway.enabled=true`, `hostnames[0]=coast.example.com` and no backend route, all three of `--set openmrs-frontend.apiUrl=`, `=openmrs` and `=./openmrs` render clean, and the SPA then calls the primary backend exactly as it did before the guard existed. The same test rejects `//api.example.com`, which is host-absolute and safe.

Keying on the scheme covers all four. Replacing lines 57-58 with:

```gotemplate
{{- $apiUrl := (index .Values "openmrs-frontend").apiUrl | default "" -}}
{{- $absoluteApi := or (hasPrefix "http://" $apiUrl) (hasPrefix "https://" $apiUrl) (hasPrefix "//" $apiUrl) -}}
{{- if and (dig "enabled" false $feGw) (not $absoluteApi) -}}
```

I ran the whole "Validate routing guard" step verbatim against that: all six of its cases still pass, the three relative forms above are now rejected, and `http://`, `https://` and `//host` still render.

.github/workflows/helm-build.yaml
suubi-joshua: Kept API-only as a valid shape. Added the NOTE at the backend-only CI test line documenting the fallthrough `/openmrs/spa` on an API-only host hits the primary's catch-all, as expected since there's no frontend route to claim it.

helm/openmrs-tenant/templates/NOTES.txt
dkayiwa: The `/etc/hosts` line lower down (line 37) still presents the Kind-only step as the general instruction. Rendering a production-shaped tenant on helm 3.17.4, with `--set "openmrs-frontend.gateway.hostnames[0]=coast.openmrs.org"`, gives:

```
To access the SPA, point the tenant hostname at the gateway first:
  echo "127.0.0.1 coast.openmrs.org" | sudo tee -a /etc/hosts
```

Nothing there marks it as local-dev, so someone installing a tenant on the shared cluster is told to point the hostname at their own machine rather than at the gateway. The heading three lines above carries the ":8080 on the Kind setup" qualifier, and the README's copy of the same step carries "(local dev)", so this is the one place it is still flat.

Not a blocker. Something like `point the tenant hostname at the gateway (DNS, or for local dev):` on line 36 would line it up with both.

dkayiwa (approved):

